PR #3, opened 2022-06-01
feat(widget-callbacks): add inline widget and callbacks from sample apps

cerebrl:


src/lib/components/primitives/button/button.story.svelte
```diff
@@ -0,0 +1,15 @@
...
+  // Button props
+  export let busy = false;
+  export let onClick = (event: Event) => {};
+  export let style: 'outline' | 'primary' | 'secondary' = 'outline';
+  export let text: string;
+  export let type: 'button' | 'submit' = null;
+  export let width: 'auto' | 'full' | 'half' = 'auto';
+</script>
+
+<Button {busy} {onClick} {style} {type} {width}>
```
ryanbas21, line 13: Is there a reason we are not using https://storybook.js.org/docs/svelte/essentials/controls would be good to use here I think for this component? It may make the story a bit clean

src/lib/components/primitives/input/floating-label.svelte
```diff
@@ -1,24 +1,34 @@
...
+  -->
   <input
+    class={`width-{width} tw_border tw_border-gray ${hasRightIcon ? 'tw_border-r-0': ''} tw_bg-white tw_block  tw_flex-1 tw_basis-auto tw_leading-6 tw_p-3 tw_rounded ${hasRightIcon ? 'tw_rounded-r-none' : ''} tw_text-base tw_text-gray-dark tw_w-1`}
     id={key}
-    {type}
     on:change={onChange}
-    class="width-{width} tw_border tw_border-gray tw_bg-white tw_block tw_leading-6 tw_mb-4 tw_p-3 tw_rounded tw_text-base tw_text-gray-dark tw_w-full"
-    {value}
     placeholder={label}
+    {type}
```
ryanbas21, line 24: I don't see any validation in the storybook when type is set to email, is this intended?

ryanbas21 (approved):
I reviewed this on chromatic, its just amazing to review there. Probably most excited i've been reviewing code. This looks really cool. I left a few comments on questions.

One thing that I'd add from the reviwers perspective is that interaction testing would be sweet because we can run the tests on the deployed storybook ourselves and see it happen.

This is really sweet though. Maybe we can look into more storybook routing integration also and link pages through it, idk it just makes reviewing the code feel like i'm actually using the app and I can't express how awesome this is.

src/lib/components/primitives/button/button.story.svelte
cerebrl: We need to discuss this as it's a bit complicated, but `*.story.svelte` is the Svelte component that's used by `*.stories.ts`. The `*.stories.ts` file is where you'll find the "controls" aspect to the story, not the component.

I've separate the component from the official `*.stories.ts` file because of this: https://github.com/storybookjs/storybook/issues/16673#issuecomment-1133477439. So, I'm avoiding using Svelte components as the official Storybook "stories" file, so we can have full functionality of Storybook. If we use the Svelte CSF (`@storybook/addon-svelte-csf`), we can't use `storyStorev7` and a bunch of other stuff, like interaction testing.

src/lib/components/primitives/input/floating-label.svelte
cerebrl: Good question. I believe validation from a native input perspective happens at form submission, not on blur or on change.

PR #6, opened 2022-07-13
feat(i18n): adding internationalization

cerebrl:
Jira ticket: https://bugster.forgerock.org/jira/browse/SDKS-1810

## Summary

This PR adds the capability of localizing content for the Widget Framework. Localized content are JSON files within the `/src/locales` directory and broken into country then language. So, for US English, the directory would be `/src/locales/us/en/index.json`.

Currently, only these four locales are supported:

- en-US
- es-US
- en-CA
- fr-CA

## Login App

The app detects the locale from the `Accept-Language` header. The locale is usually in this format: `en-US`, but it can also just be a language: `en`. So, there's a utility to help determine what locale content to load. If country is not specified, then US is used.

## Login Widget

The widget doesn't detect locale, but is just given the content from the parent application. This allows the parent application to have 100% control over how it determines the locale and what content to render. The localized content is passed in via props when the widget is initialized.

TODO:

- [x] Add e2e testing for localization with Playwright
- [x] Add unit tests for locale/language utility functions
- [x] Add e2e test for handling for just language without country


cerebrl:
> Not sure why I don't see UI review links above - is this expected? Also the preview app doesn't seem to work - looks like it's failing with a CORS error in the authenticate call.

Yes, the Chromatic and Storybook artifacts don't get deployed unless we can successful CI. The preview also requires a manual addition of the automated Vercel URL to my tenant for CORS.

With that said, this is a work in progress, so you can ignore for now.

ryanbas21:
> The preview also requires a manual addition of the automated Vercel URL to my tenant for CORS.

We can use my github action if we want to here.

cerebrl:
Added more tests and fixed a few tiny bugs.

src/lib/components/i18n/index.svelte
```diff
@@ -0,0 +1,15 @@
...
+
+  export let html = false;
+  export let key: string;
+  export let values: Record<string, string> | undefined = undefined
+
+  let message = interpolate(key, values);
+</script>
+
+{#if html}
+  <slot>{@html message}</slot>
```
cerebrl, line 12: This is the main reason for this component. The `@html` is the same as `dangerouslySetInnerHtml` in React. Since it's a liability, I wanted to do it in one place, and only after I know the contents have been cleaned by the `interpolate` function.

src/lib/journey/callbacks/boolean/boolean.svelte
```diff
@@ -37,11 +40,10 @@
...
 <Checkbox
   errorMessage={validationFailure}
-  {isRequired}
   isInvalid={!!validationFailure}
   key={inputName}
   onChange={setValue}
   value={previousValue}
 >
-  {prompt}
+  {interpolate(outputName.replace('/', ''), null, prompt)}
```
cerebrl, line 48: You'll notice that all the changes to the callback handlers is essentially the use of this `interpolate` function. Which replaces hard coded strings.

src/lib/journey/utilities/callback.utilities.ts
```diff
@@ -5,7 +5,12 @@ import type {
   ValidatedCreatePasswordCallback,
 } from '@forgerock/javascript-sdk';
 
-type RequiredCallbacks = AttributeInputCallback<boolean | string> | ValidatedCreatePasswordCallback | ValidatedCreateUsernameCallback;
+import { interpolate } from '$lib/utilities/i18n.utilities';
```
cerebrl, line 8: Ensuring that all error messages from the callbacks come from translated text.

src/lib/locale.store.ts
```diff
@@ -0,0 +1,63 @@
...
+  securityQuestionsPrompt: z.string(),
+  showPassword: z.string(),
+  sn: z.string(),
+  submitButton: z.string(),
+  successMessage: z.string(),
+  termsAndConditions: z.string(),
+  useValidEmail: z.string(),
+  ValidatedCreatePasswordCallback: z.string(),
+  ValidatedCreateUsernameCallback: z.string(),
+});
```
cerebrl, line 39: This is Zod. You can check it out here: https://zod.dev. Let me know what you think about it. I really like its use for runtime object schema validation and its ability to generate TypeScript types from these schemas. This will be really important when this eventually gets used as a framework. It will give the devs immediate feedback on config objects, content objects ... anything that has an important schema.

src/lib/locale.store.ts
```diff
@@ -0,0 +1,63 @@
...
+const partialStringsSchema = stringsSchema.partial();
+
+// Ensure fallback follows schema
+stringsSchema.parse(fallback);
+
+export const locale: Writable<string | null> = writable('en-US');
+export let strings: Readable<Record<string, string> | null>;
+
+export function initializeContent(userLocale?: z.infer<typeof partialStringsSchema>, partial?: boolean) {
```
cerebrl, line 48: This is setting the content in a store, so it's available to the components.

src/lib/utilities/i18n.utilities.ts
```diff
@@ -0,0 +1,115 @@
...
+import { strings } from '$lib/locale.store';
+
+/**
+ * Do not allow strings with angle brackets, just to be extra safe
+ *
+ * Demo: https://regex101.com/r/Mw9vTB/1
+ */
+const valueSchema = z.record(z.string().regex(/^[^<>]*$/)).optional();
+
+export function getLocale(acceptLanguageHeader: string, delimiter: '/' | '_'): string {
```
cerebrl, line 14: Converts locales to directory paths or object paths.

src/lib/widget/modal.svelte
```diff
@@ -85,6 +86,8 @@
   let _dialogEl: HTMLDialogElement;
   let _journeyComp: SvelteComponent;
 
+  initializeContent(content, true);
```
cerebrl, line 89: For the Widget (both inline and modal), we essentially take the content provided by the parent app and initialize the content store with it.

src/routes/api/locale.ts
```diff
@@ -0,0 +1,27 @@
...
+import us_es from '$locales/us/es/index.json';
+
+const locales = {
+  ca_en,
+  ca_fr,
+  us_en,
+  us_es,
+}
+
+export async function get(event: RequestEvent) {
```
cerebrl, line 18: This is just an API endpoint that can be used for fetching localized content, but it's not really used right now.

src/routes/e2e/widget/inline.svelte
```diff
@@ -17,21 +17,30 @@
...
-          redirectUri: 'https://localhost:3000/callback',
+          redirectUri: `${window.location.origin}/callback`,
           scope: 'openid profile email me.read',
           serverConfig: {
             baseUrl: 'https://openam-crbrl-01.forgeblocks.com/am/'
           },
           realmPath: 'alpha',
           tree: 'Login',
         },
+        content,
```
cerebrl, line 43: This is the content object passed into the Widget initialization. This is where we are using the locale API endpoint.

src/routes/index.svelte
```diff
@@ -20,12 +20,17 @@
 
   import Journey from '../lib/journey/journey.svelte';
   import { initialize, type InitObject } from '$journey/journey.store';
+  import { initializeContent } from '$lib/locale.store';
+
+  export let content;
```
cerebrl, line 25: This `content` is provided by the "shadow endpoint" that matches this page component, for this, that would be `./index.ts`.

src/routes/index.ts
```diff
@@ -0,0 +1,28 @@
...
+import us_es from '$locales/us/es/index.json';
+
+const locales = {
+  ca_en,
+  ca_fr,
+  us_en,
+  us_es,
+}
+
+export async function get(event: RequestEvent) {
```
cerebrl, line 18: This is the `index.svelte`'s "shadow endpoint". When the page is rendered on the server, the page component calls this function directly. When loaded on the client, it makes an HTTP Get call for this content.

src/routes/index.ts
```diff
@@ -0,0 +1,28 @@
...
+  const userLocale = event.request.headers.get('accept-language') || 'en-US';
+
+  const locale = getLocale(userLocale, '_');
+
+  return {
+    body: {
+      content: locales[locale],
```
cerebrl, line 25: This gets mapped to the `export let content` in the `index.svelte` page component.

src/lib/journey/callbacks/kba/kba-create.svelte
```diff
@@ -18,12 +20,13 @@
   const prompt = callback.getPrompt();
   const questions = callback
     .getPredefinedQuestions()
-    ?.map((label, idx) => ({ text: label, value: idx }));
+    ?.map((label, idx) => ({ text: label, value: `${idx}` }));
```
ryanbas21, line 23: can `predefinedQuestions` be null or undefined? should we then do an optional chain on the`getPrompt` call? I feel like if we succeed the `prompt` callback call, then we would have predefined questions?

src/lib/utilities/i18n.utilities.ts
```diff
@@ -0,0 +1,115 @@
...
+    messageDirty = textFromKey.charAt(0).toUpperCase() + textFromKey.slice(1);
+  } else if (!messageDirty && externalText) {
+    /**
+     * If no internal message is found, but external content is provided,
+     * just return the external text
+     */
+    messageDirty = externalText;
+  }
+
+  const messageClean = xss(messageDirty);
```
ryanbas21, line 112: `sanitize` feels like a more common word for this, no?

tsconfig.json
```diff
@@ -1,16 +1,20 @@
 {
   "extends": "./.svelte-kit/tsconfig.json",
   "compilerOptions": {
+    "allowSyntheticDefaultImports": true,
```
ryanbas21, line 4: Why do we want this? Typically this is solved by doing `* as object` because the `export` is not an actual default module.

Is this because a new library is using a commonjs export and not a `export default`? I'm guessing this was a debug of why `xss` wasn't importing correctly?

src/lib/utilities/i18n.utilities.ts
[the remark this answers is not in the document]
ryanbas21: messageDirty feels like it should be sanitize, but this is just how i've learned the concept.

`messageDirty` to me reads like a form state being dirty. I'm not sure if i'm just weird but i've always seen `dirty` refer to a `state` and `sanitize` refer to the "cleaning of an input" which is kind of whats happening here.

ryanbas21 (approved):
The svelte/framework code looks good, i just have a few outstanding comments/questions. I was able to track this and follow pretty easily i felt like.

I probably need to _use_ zod to have more of an opinon, but the concept seems similar to io-ts which does similar safety.

src/lib/components/compositions/checkbox/animated.story.svelte
```diff
@@ -1,14 +1,47 @@
...
   export let key: string;
   export let label: string;
   export let onChange: (event: Event) => void;
   export let value: boolean;
+  export let withForm = false;
+
+  let el;
+
+  function submitForm(event: SubmitEvent) {
+    console.log('Form submitted');
```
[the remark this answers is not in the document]
cerebrl: Since this is early in development, I'm okay with console logs. But yes, once this gets closer to production, we'll clean all these console logs out.

src/lib/journey/callbacks/kba/kba-create.svelte
cerebrl: `callback` will exist within this section of code, but there's not a guarantee of any predefined questions. There's a possibility of this callback only having custom questions. This is where the user defines their own question _and_ answer, rather than being given an existing question to answer.

It's worth noting that at this stage of development, none of the callback handler components are actually "done". The current story I'm working on will further develop the callback handlers associated with the default login tree. Then, there's another story that does the same for the default registration.

The current callback handler components you see here are developed just enough for testing the happy path.

src/lib/utilities/i18n.utilities.ts
cerebrl: Are you suggesting to call the default import on the `xss` module `sanitize`? If so, I can absolutely do that. All the docs for this module use `xss`, so I just copied that over. But you're right; using `xss` doesn't make a ton of sense.

tsconfig.json
cerebrl: This was to allow the import of JSON files, as they don't have a default export. Do you know of another way to import JSON files without using this config setting?

src/lib/utilities/i18n.utilities.ts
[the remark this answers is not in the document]
cerebrl: Is this comment also related to the use of `xss` as the function name? I'm assuming you're not wanting to change the nouns (`messageDirty` and `messageClean`), right?

ryanbas21:
```suggestion
  const sanitize = xss(messageDirty);
```
```suggestion
  const sanitizedMessage = xss(messageDirty);
```
```suggestion
const sanitizedInput = xss(messageDirty);
```

tsconfig.json
ryanbas21: I think all that would be needed is  `resolveJsonModule: true`

https://www.typescriptlang.org/tsconfig#resolveJsonModule

src/locales/ca/en/index.json
```diff
@@ -0,0 +1,34 @@
...
+  "securityQuestions": "Security question(s)",
+  "securityQuestionsPrompt": "Provide security question(s) & answer(s) below",
+  "showPassword": "Show Password",
+  "sn": "Last Name",
+  "submitButton": "Submit",
+  "successMessage": "Success!",
+  "termsAndConditions": "Please accept our Terms and Conditions",
+  "useValidEmail": "Please use a valid email address.",
+  "ValidatedCreatePasswordCallback": "Password",
+  "ValidatedCreateUsernameCallback": "Username"
```
daveadams56, line 33: I was a little confused about what was being defined here at first - I can see some of these keys seem to correspond to the type of the callback, and others seem to be matching to what I think is probably the prompt for a callback, set in AM. Others seem to be keys created within the code based on how you are processing the validation policy failures. I guess my concern is that as an app and the flows implemented on AM grow, this could become quite hard to manage given the mix of things that are coupled to AM prompts, things that are not and things that only exist within this app. What would you think about structuring the json files in such a way as they separate "coupled FR stuff" from "non coupled FR stuff" e.g. callback types, and "app strings"?

src/lib/journey/callbacks/boolean/boolean.svelte
daveadams56: I left a comment on [src/locales/ca/en/index.json](https://github.com/cerebrl/forgerock-web-login-framework/pull/6/files/2a6df6259179921f4523db6c4f9c33c287df8736#diff-0b04e3700cfa2909235f08372cab26546edbb56b662609ac31e76de730052a6a) as it is still a little unclear how some of these strings are being used, e.g. some seem to be callback types, some seem to be based on the prompts configured in AM. Where are the callback type names being interpolated?

src/lib/journey/callbacks/kba/kba-create.svelte
```diff
@@ -81,7 +84,7 @@
 
   <Input
     key={inputNameAnswer || 'ka-answer-label'}
-    label="Security Answer"
+    label={interpolate("securityAnswer")}
```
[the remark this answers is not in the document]
daveadams56: Additionally, I think IDM allows localization of the KBA answers within the product itself. Are we leveraging any of this, or is there just a "en" answer in IDM and we localize it at the app?

src/lib/journey/stages/generic.svelte
```diff
@@ -1,34 +1,37 @@
...
-  <p>Loading ...</p>
-{:else if $step.type === 'Step'}
+{#if !step}
+  <div class="tw_text-center tw_w-full tw_py-4">
+    <Spinner colorClass="tw_text-primary-light" layoutClasses="tw_h-28 tw_w-28" />
+  </div>
+{:else if step.type === 'Step'}
+  <h1 class="tw_primary-header dark:tw_primary-header_dark">
+    <!-- TODO: Needs localization strategy -->
+    {step.getHeader() || ''}
```
daveadams56, line 25: Are we allowing HTML to be rendered from within the header as defined in AM? I have seen this before, I think even on default journeys. Not sure if it's a convention or expectation that a UI will render this as platform-login does. Perhaps having this as an option for the app developer to turn on / off could be good, when we do implement localization for this bit.

src/lib/components/i18n/index.svelte
daveadams56: Ah so this is how we might let for example the step.getHeader() text render HTML - once it has been sanitized? I think I get this now

src/lib/journey/callbacks/terms-and-conditions/terms-conditions.svelte
```diff
@@ -34,6 +34,5 @@
 </script>
 
 <Checkbox key={inputName} onChange={setValue} value={false}>
-  <!-- TODO: Remove hardcoded text below -->
-  Please accept our below Terms and Conditions
+  <T key="termsAndConditions" />
```
daveadams56, line 37: I don't see the terms and conditions rendered for this component. Guessing this has not been implemented yet? Also on this point what is our strategy for terms and conditions localization given IDM has its own mechanism for this? (similar to KBA answers)?

src/routes/register.ts
```diff
@@ -0,0 +1,28 @@
...
+import { getLocale } from '$lib/utilities/i18n.utilities';
+
+// Supported locales
+import ca_en from '$locales/ca/en/index.json';
+import ca_fr from '$locales/ca/fr/index.json';
+import us_en from '$locales/us/en/index.json';
+import us_es from '$locales/us/es/index.json';
+
+const locales = {
+  ca_en,
```
daveadams56, line 12: I was slightly surprised to see the locales hard coded within the app - might this increase the "cost" of adding more locales later once the app is in production? My thinking is of the effort needed to support a new language with a code and config change, as opposed to just a config change e.g. drop in a new language file with a filename matching the locale. What are your thoughts about this?

src/lib/utilities/i18n.utilities.ts
```diff
@@ -0,0 +1,115 @@
...
+    splitResult[0].length > 4
+  ) {
+    console.warn('Locale from Accept-Language header not recognized');
+    return `us${delimiter}en`;
+  }
+
+  return `${splitResult[1] || 'us'}${delimiter}${splitResult[0]}`.toLowerCase();
+}
+
+export function interpolate(
```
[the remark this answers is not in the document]
daveadams56: I take it that in most / all instances, we will use the localized text included in the app, over a value from AM? I'm keen to discuss further in more detail, the overall strategy for localization between the app and platform, i.e. if we are consciously not using some platform capabilities, is what we are doing here a pattern that all customers should be following? Or some - and where is the line between the app and the platform's responsibility?

src/lib/journey/callbacks/username/validated-create-username.svelte
```diff
@@ -33,13 +35,17 @@
...
     callback.setInputValue((event.target as HTMLInputElement).value);
   }
+
+  $: {
+    console.log(callback);
+  }
 </script>
 
 <Input
   errorMessage={validationFailure}
```
daveadams56, line 45: I saw some weird behaviour on this field - the message "Please check this value" was repeated four times, presumably because there are four (different) validation failures, e.g:

"{ \"policyRequirement\": \"REQUIRED\" }",
"{ \"params\": { \"minLength\": 1 }, \"policyRequirement\": \"MIN_LENGTH\" }",
"{ \"params\": { \"minLength\": 1 }, \"policyRequirement\": \"MIN_LENGTH\" }",
"{ \"params\": { \"maxLength\": 255 }, \"policyRequirement\": \"MAX_LENGTH\" }"

src/locales/ca/en/index.json
```diff
@@ -0,0 +1,34 @@
...
+  "dontHaveAnAccount": "Don’t have an account? <a href='/register'>Sign up here!</a>",
+  "closeModal": "Close Modal",
+  "chooseDifferentUsername": "Please choose a different username.",
+  "ensurePasswordIsMoreThan": "Ensure password contains more than {minPasswordLength} characters.",
+  "ensurePasswordHasOne": "Ensure password contains 1 of each: capital letter, number and special character.",
+  "givenName": "First Name",
+  "loading": "Loading ...",
+  "loginButton": "Login In",
+  "loginHeader": "Login In",
+  "loginSuccess": "Login Success!",
```
daveadams56, line 12: We have a message for login success, but nothing for login failure. The app does not seem to render any form of failure page, instead rendering the login page again. Is there a better way we can communicate to the user what the problem might be on login?

src/routes/api/locale.ts
daveadams56: Is this exclusively for E2E tests, or did you mean it is used by E2E as well as the app as a whole?

src/routes/e2e/widget/modal.svelte
```diff
@@ -28,26 +28,38 @@
...
+    }
+
     widget = new Widget({
       target: widgetEl,
       props: {
         config: {
           clientId: 'WebOAuthClient',
-          redirectUri: 'https://localhost:3000/callback',
+          redirectUri: `${window.location.origin}/callback`,
           scope: 'openid profile email me.read',
```
daveadams56, line 47: Might it be worth making the scopes configurable?

src/lib/widget/modal.svelte
daveadams56: I like the way this keeps the widget "pure" and leaving those concerns to the parent app. How do we plan to communicate the expectations / assumptions on this content, to developers who implement the widget in their app?

daveadams56: e.g. are we going to provide sample translation files with the widget with a complete list of the strings that are required for the callbacks to work, and their counterpart strings built into the AM trees?

src/locales/ca/en/index.json
cerebrl: This is what I brought up in the demo two Fridays ago. When it comes to callbacks, localization is very difficult. Callbacks rarely have something to "key off of" for localized content. So, I did my best and used what was available to me.

I guess I could place a bit more responsibility in the callback handler to generate a more "normalized" key for the content, avoiding the rather heterogenous keys like what you see here. But, I'm open to suggestions.

daveadams56 (commented):
Overall this is looking great and I am starting to understand the decisions that have had to be made here - and I get that there may be no ideal or one way to localize certain things that are generated from the server. None of the issues or comments I've left are blocking especially given the stage of development this is at. I'm impressed at how solid this is looking, to be honest! One piece of feedback (and may be something you've answered), is about the review process on Chromatic given the diff does not appear for the whole change set, and requires you to go commit by commit. Am I missing a better way to do this?

src/lib/journey/stages/generic.svelte
cerebrl: The challenge here is that many of these strings that can be set in AM, like this header value, don't provide alternative language values. But yes, the values of `header` and `description` can be regular strings or HTML, but in situations where localization matters, these values are hardcoded to a single language, so they can't be used.

My suggestion would be to just use the `header` and `description` value coming from AM (page node to be exact) as the key itself. So, rather than the `header` value being `"Sign Up, Today!"`, for example, it would be `"signUpToday"`.

> Perhaps having this as an option for the app developer to turn on / off could be good

We might be able to do this, but I think it's a different user story. The downside to this is we lose the single source of truth for content. Some things have to be provided by the content files; some things that come from AM don't have prompts or labels, some things do but only in a single language, some things have multiple languages ... it's just keeping track of all of that in a resilient way.

I think we might be able to come to a better conclusion once I can speak with Mike.

daveadams56:
@ryanbas21 I commented on one of the tests: [Terms and Conditions](https://www.chromatic.com/test?appId=628d1f65487489003ae0890d&id=62d96fc16faff6c246c60a93) and resolved the conversation as I didn't see it as a blocker to being approved. Are you happy for me to keep that one resolved - and resolve this one? [ValidatedCreateUsername](https://www.chromatic.com/test?appId=628d1f65487489003ae0890d&id=62d96fc16faff6c246c60a96)

src/lib/journey/callbacks/terms-and-conditions/terms-conditions.svelte
cerebrl: Oh, and yes, this is very much like the KBA in the fact that some things will come "pre-localized" (if you will) from AM/IDM. This is one of them.

src/routes/register.ts
cerebrl: I was originally using purely directory/file based methods, but I ended up being completely obstructed by the affect this had in deployment. The locales weren't being deployed because nothing was referencing them in the app. E.g.: I was using an `async import` like this: 

```
async import(`locales/${localDir}/index.json`);
```

The problem is when the above gets compiled for deployment, all of the local files are not "statically" referenced, so they are not deployed. If I can figure out how to get around that, then we can revisit it. But, to no block the rest of the project, I needed a way to get around it and move on.

src/locales/ca/en/index.json
cerebrl: Step submission failures are not yet handled within the app. That's what I'm working on this sprint, so I'll add any localizations to the app through that work.

src/lib/components/i18n/index.svelte
cerebrl: This is definitely possible, but it's an open question about how to handle localization from a system perspective. I'd, personally, recommend all content come from the localization files within the project as this lines up better with how large company's handle this organizationally. That way, you have one source of truth that can be updated and controlled by the same team.

Now, there are instances where this isn't 100% possible, and some localized content does come from AM/IDM. This is not ideal, but it's just something we have to deal with. When it comes to choices, like page node headers and descriptions, I'd recommend they use a single value in AM as a key to lookup the content from within the app, even though it _could_ be localized in AM.

I say that because most big companies have a localization team that's responsible for maintaining the locale files for the user facing apps. UI engineers rarely worry about these files, and they just provide the US English strings provided by the design or content team. The localization team then takes the provided content and adds the new keys and translated strings to the supported locale files. They are not likely to have access or want to rummage through AM for translated strings.

Does that make sense? This topic gets complicated and deep very quickly.

src/routes/register.ts
cerebrl: If we can revisit the ability to use `async imports`, I too would prefer using that over this approach.

tsconfig.json
cerebrl: Hmm, okay, I'll have to try that.

daveadams56 (approved):

PR #9, opened 2022-07-31
feat(login-logout): complete login, oauth flow and logout

cerebrl:
Jira ticket: https://bugster.forgerock.org/jira/browse/SDKS-1811

## Summary

Build out the complete functionality of the Login flow: journey, OAuth, user info as well as complete logout.

## Details

Major refactor of `journey.store.ts` and `journey.svelte`. Much of the logic and orchestration has been moved up into the widget's top-level files. This is for improved configuration and direct control, rather than being buried deep in a component some where.

Another major change is the use of custom Svelte stores, rather than plain objects that contain multiple stores. This helps encapsulate/group things together in the same store and reduce complexity. All stores will share a common interface as well.

Directory structure changed just a touch. The `widget` directory is now out of the `lib` directory to help differentiate generic reusable code from top-level app code.

## Testing

Please use these URLs for testing the widget: 

- https://forgerock-web-login-framework-git-featlogin-flow-cix.vercel.app/e2e/widget/modal
- https://forgerock-web-login-framework-git-featlogin-flow-cix.vercel.app/e2e/widget/inline


.storybook/main.js
```diff
@@ -53,7 +53,7 @@ module.exports = {
           $journey: resolve('./src/lib/journey'),
           $lib: resolve('./src/lib'),
           $locales: resolve('./src/locales'),
-          $widget: resolve('./src/lib/widget'),
+          $widget: resolve('./src/widget'),
```
cerebrl, line 56: Moved the `widget` directory out of the `lib` directory as I'm considering anything in `lib` to be a reusable component, but the top-level widget stuff is the entry file into the app, which isn't a reusable thing ... kind of like the routes.

src/lib/components/compositions/checkbox/animated.svelte
```diff
@@ -1,22 +1,33 @@
...
     }
     onChange(event);
   }
 
+  afterUpdate(() => {
+    if (firstInvalidInput) {
+      inputEl.focus();
+    }
+  });
+
```
cerebrl, line 30: You'll see these changes repeated throughout the components. It's what supports keyboard focus management after network request, as opposed to what you saw before, which was focus management before network request.

src/lib/components/primitives/alert/alert.svelte
```diff
@@ -0,0 +1,49 @@
+<script lang="ts">
```
cerebrl, line 1: This is the only new component being added. This needs some design love, but leaving as is to keep moving forward.

src/lib/components/primitives/button/button.story.svelte
```diff
@@ -1,18 +1,18 @@
 <script lang="ts">
   import Button from "./button.svelte";
-  import { initStyles } from '$widget/styles.store';
+  import { initialize } from '$widget/styles.store';
 
   // Button props
   export let busy = false;
-  export let customCss = [];
+  // export let customCss = [];
```
cerebrl, line 7: Throughout this PR, you'll notice I'm deactivating the "runtime" customization stuff for now. We'll have a future story to implement this more maturely.

src/lib/components/primitives/form/form.svelte
```diff
@@ -29,38 +29,47 @@
     // Iterate over all children of the form, and pluck out the the inputs
```
cerebrl, line 29: This is the form component that gives us the capability to do client-side validation. So, if AM provides metadata on callbacks informing us that they are required or have min/max values ... whatever, we can implement it using the native DOM validation. Whatever slips through will be caught be AM and that validation will be handled separately.

src/lib/journey/callbacks/password/validated-create-password.svelte
```diff
@@ -5,11 +5,26 @@
   import { getPasswordValidationFailureText, isInputRequired } from '$journey/utilities/callback.utilities';
 
   export let callback: ValidatedCreatePasswordCallback;
+  export let firstInvalidInput: boolean;
   export let idx: number;
 
+  /**
+   * This callback does not indicate that the input is required.
+   * TODO: Can we just assume a password is always required?
+   */
```
cerebrl, line 14: Sorry, I need to remove this comment. The answer is no, we can't assume password is always required. The below `isInputRequired` will always return false for this callback. The only policy on the callback is the following:

```
policyRequirements: ["VALID_TYPE"]
```

`"VALID_TYPE"` is not very informative, and there doesn't seem to be a way to mark it as required in the Admin UI 🤷  So, I guess this will always slip through and the server will have to validate.

src/lib/journey/callbacks/string-attribute/string-attribute-input.svelte
```diff
@@ -35,10 +36,21 @@
...
+
+  $: {
+    isRequired = isInputRequired(callback);
+    outputName = callback.getOutputByName('name', '');
+    policies = callback.getPolicies();
+    type = getInputTypeFromPolicies(policies);
+    previousValue = callback?.getInputValue() as string;
+    textInputLabel = callback.getPrompt();
+    validationFailure = getAttributeValidationFailureText(callback);
+  }
```
cerebrl, line 48: It's worth noting that with Svelte, if you have a local variable that is assigned a value through the calling of a function, you have to wrap those assignments in a reactive block. This is because Svelte won't rerun all function within a component, so you need to explicitly tell Svelte to rerun the ones you want.

Ordinary, direct, variable assignment is observed by default, so it's just assignment through functions to be mindful of.

src/lib/journey/callbacks/terms-and-conditions/terms-conditions.svelte
```diff
@@ -5,8 +5,11 @@
   import T from '$components/i18n/index.svelte';
 
   export let callback: TermsAndConditionsCallback;
+  export let firstInvalidInput: boolean;
   export let idx: number;
 
+  // TODO: Component needs a UX story to be complete
```
cerebrl, line 11: I'll write up a story for this component.

src/lib/journey/journey.store.ts
```diff
@@ -4,72 +4,57 @@ import {
   FRLoginFailure,
   FRLoginSuccess,
   StepType,
-  TokenManager,
-  UserManager,
 } from '@forgerock/javascript-sdk';
-import { get, writable, type Writable } from 'svelte/store';
+import type { Step, StepOptions } from '@forgerock/javascript-sdk/lib/auth/interfaces';
```
cerebrl, line 8: This is the component that had the heaviest refactoring.

src/lib/journey/journey.store.ts
```diff
@@ -4,72 +4,57 @@ import {
...
-import type { StringDict } from '$lib/interfaces';
 import { htmlDecode } from '$lib/journey/utilities/decode.utilities';
-import { email, isAuthenticated, fullName } from '$lib/user/user.store';
 
-export interface InitObject {
-  step: Writable<StepTypes>;
-  getStep: (prevStep?: StepTypes) => Promise<void>;
-  failureMessage: Writable<string | null>;
-  submittingForm: Writable<boolean>;
+export interface JourneyStore extends Pick<Writable<JourneyStoreValue>, 'subscribe'> {
```
cerebrl, line 13: I converted the export from a regular function to a custom store (observable). This helps leverage Svelte's own features and normalizes the stores' interfaces.

src/lib/journey/journey.store.ts
```diff
@@ -4,72 +4,57 @@ import {
...
+  // TODO: Think about turning this into an object with code, message and step
+  error: {
+    code: number | null;
+    message: string | null;
+    step: Step | undefined;
+  } | null;
+  loading: boolean;
+  step: StepTypes;
+  successful: boolean;
+  response: Step | null | undefined;
```
cerebrl, line 28: This is the standard interface for all stores that return to the application developer.

src/lib/journey/journey.store.ts
```diff
@@ -84,52 +69,72 @@ export async function initialize(journey?: string): Promise<InitObject> {
...
-        message: 'Unknown request failure'
+        message: 'Unknown request failure',
       });
     }
 
     if (nextStep.type === StepType.Step) {
-      step.set(nextStep);
-      submittingForm.set(false);
+      // Iterate on a successful progression
+      stepNumber = stepNumber + 1;
```
cerebrl, line 99: This is just for tracking when we can autosubmit the form if the form data got stale. Step #1 is the only valid number to resubmit.

src/lib/journey/journey.store.ts
```diff
@@ -4,72 +4,57 @@ import {
...
+  tokenId: string;
 }
 export type StepTypes = FRStep | FRLoginSuccess | FRLoginFailure | null;
 
-export async function initialize(journey?: string): Promise<InitObject> {
-  const step: Writable<StepTypes> = writable(null);
-  const failureMessage: Writable<string | null> = writable(null);
-  const options: Options = {};
-  const submittingForm: Writable<boolean> = writable(false);
+export function initialize(initOptions?: StepOptions): JourneyStore {
```
cerebrl, line 35: This just initializes the journey store and provides a closure for configuration values. It does not call AM, just sets everything up.

src/lib/journey/journey.store.ts
```diff
@@ -84,52 +69,72 @@ export async function initialize(journey?: string): Promise<InitObject> {
...
+      successful: false,
+      response: null,
+    });
+
     try {
       /**
        * Initial attempt to retrieve next step
        */
-      nextStep = await FRAuth.next(get(step as Writable<FRStep>), options);
+      nextStep = await FRAuth.next(prevStep as FRStep, options);
```
cerebrl, line 85: This is the first call to AM. It could be with or without a previous step.

src/lib/journey/journey.store.ts
```diff
@@ -84,52 +69,72 @@ export async function initialize(journey?: string): Promise<InitObject> {
...
     } catch (err) {
       console.error(`Next step request | ${err}`);
 
       /**
        * Setup an object to display failure message
        */
       nextStep = new FRLoginFailure({
-        message: 'Unknown request failure'
+        message: 'Unknown request failure',
```
cerebrl, line 93: Ah, just realized this is not localized! Oops.

daveadams56 (commented):
I think the preview app may be misconfigured; I can log in and see the "success" banner, however the OAuth flow does not succeed. I think this is because the **/authorize** call returns a redirect to https://localhost:3000/callback instead of current app URL.

src/lib/journey/journey.store.ts
```diff
@@ -139,32 +144,113 @@ export async function initialize(journey?: string): Promise<InitObject> {
...
+                callback.setInputValue('');
+                return callback;
+              },
+            );
+            restartedStep.payload = {
+              ...previousPayload,
+              authId: restartedStep.payload.authId,
+            };
+          }
+        }
```
cerebrl, line 169: All of this is going to be removed. I think this is legacy code that I added a long time ago due to a bad ID Cloud tenant.

src/lib/journey/journey.store.ts
```diff
@@ -139,32 +144,113 @@ export async function initialize(journey?: string): Promise<InitObject> {
...
+              restartedStep.callbacks = previousCallbacks;
+              restartedStep.payload = {
+                ...previousPayload,
+                authId: restartedStep.payload.authId,
+              };
+            }
+            restartedStep = await FRAuth.next(restartedStep, options);
+          }
+        }
+      }
```
cerebrl, line 193: This is the autosubmission if the reason for the failure is `authId` timeout.

src/lib/journey/journey.svelte
```diff
@@ -1,97 +1,41 @@
 <script lang="ts">
```
cerebrl, line 1: This component also got heavily refactored/simplified with the `journey.store.ts` changes.

src/lib/journey/stages/generic.svelte
```diff
@@ -1,37 +1,69 @@
...
+  function checkValidation(callback: any) {
+    let failedPolices = callback.getOutputByName('failedPolicies', []);
+    if (failedPolices.length && !hasPrevError) {
+      console.log(callback);
+      hasPrevError = true;
+      return true;
+    }
+    return false;
+  }
+
```
cerebrl, line 36: You may see a few of these comments saying things should be moved to utility functions. There's some duplication of logic across components. The fact that there are some closures dependencies makes it a bit harder to move them out.

src/lib/oauth/oauth.store.ts
```diff
@@ -0,0 +1,86 @@
...
+}
+export interface OAuthTokenStoreValue {
+  completed: boolean;
+  error: {
+    code?: number | null;
+    message: string | null;
+  } | null;
+  loading: boolean;
+  successful: boolean;
+  response: OAuth2Tokens | null | void;
```
cerebrl, line 20: Here's that standard interface I mentioned above.

src/routes/e2e/widget/inline.svelte
```diff
@@ -1,20 +1,18 @@
...
 
   async function logout() {
     await user.logout();
-    userInfo = null;
+    userResponse = null;
   }
 
   form.onMount((component: HTMLElement) => console.log(component));
-  journey.onSuccess((response: User) => (userInfo = response));
+  journey.onSuccess((response: any) => (userResponse = response?.user));
```
cerebrl, line 15: This is change one of two from the usage of the Widget standpoint.

src/routes/e2e/widget/modal.svelte
```diff
@@ -1,27 +1,26 @@
...
   }
 
-  modal.onMount((dialog: HTMLDialogElement) => {
-    console.log('Singleton onMount event fired');
+  modal.onMount((dialog: HTMLDialogElement, form: HTMLFormElement) => {
     console.log(dialog);
+    console.log(form);
 
-    journey.initialize();
+    journey.start();
```
cerebrl, line 19: Calling initialize is no longer needed, they just need to call start.

src/widget/inline.svelte
```diff
@@ -0,0 +1,230 @@
...
+  }
+
+  let callMounted: (form: HTMLFormElement) => void;
+  let journeyStore: JourneyStore;
+  let oauthStore: OAuthStore;
+  let returnError: (response: Response) => void;
+  let returnResponse: (response: Response) => void;
+  let userStore: UserStore;
+
+  export const journey = {
```
cerebrl, line 31: This portion of the component kind of blew up in size. So much of this is duplicated between the modal and inline component, but I'm ensure how to disentangle them, atm. Will do it in a future story.

src/widget/modal.svelte
```diff
@@ -0,0 +1,277 @@
...
+  let dialogComp: SvelteComponent;
+  let dialogEl: HTMLDialogElement;
+  let callMounted: (dialog: HTMLDialogElement, form: HTMLFormElement) => void;
+  let journeyStore: JourneyStore;
+  let oauthStore: OAuthStore;
+  let returnError: (response: Response) => void;
+  let returnResponse: (response: Response) => void;
+  let userStore: UserStore;
+
+  export const journey = {
```
cerebrl, line 42: Again, this is huge and duplicated, but I just didn't have the time to figure out how to break this out without breakage.

ryanbas21 (commented):
Initial pass: 

I notice the login modal is loaded on page load rather than on button press, this could improve the TTI of the base page

We also load the `error` bundle as part of the modal, even if there is no error, I think this could also improve page scores if we only load JS that is necessary.

I know this is an "e2e" app, but this is how I would expect/instruct/hope someone uses a modal (loading it when required not always) so we should probably test it as such.


I'd like to highlight the great caching work being done (by svelte under the hood). 

Screen Shot 2022-08-01 at 11 21 51 AM


Do you know why we load a `filename.svelte` and a `filename` bundle? I couldn't figure out exactly the difference other than i think the svelte is a component bundle?

src/lib/components/compositions/checkbox/animated.svelte
daveadams56: Makes sense, this works nicely on the registration page if testing in the app

daveadams56: Actually looks like I was talking about text input fields after network request, not sure how to test checkbox but I'm sure this makes sense

src/lib/components/primitives/alert/alert.svelte
daveadams56: Looks good so far

src/lib/journey/callbacks/kba/kba-create.svelte
```diff
@@ -57,6 +60,12 @@
      *********************************************************************** */
     callback.setQuestion((event.target as HTMLSelectElement).value);
   }
+
+  $: {
+    if (callback) {
+      value = '';
```
daveadams56, line 66: Curious, why are we doing this just for KBA create and not any of the other callbacks?

src/lib/journey/callbacks/password/validated-create-password.svelte
daveadams56: Do we have plans to utilise the "real time" validation of the password the user is creating, as the Platform Login UI does? This seems to submit the form using the **validateOnly** flag. I think this is a nice feature to include but I don't consider it blocking to creation of the alpha version. @cerebrl thoughts?

src/lib/components/primitives/form/form.svelte
daveadams56: Think this makes sense

src/lib/journey/callbacks/string-attribute/string-attribute-input.svelte
```diff
@@ -10,16 +10,17 @@
...
-  const type = getInputTypeFromPolicies(policies);
-  const previousValue = callback?.getInputValue() as string;
-  const textInputLabel = callback.getPrompt();
-  const validationFailure = getAttributeValidationFailureText(callback);
+  let inputName = callback?.payload?.input?.[0].name || `string-attr-${idx}`;
+  let isRequired = isInputRequired(callback);
+  let outputName = callback.getOutputByName('name', '');
+  let policies = callback.getPolicies();
+  let type = getInputTypeFromPolicies(policies);
+  let previousValue = callback?.getInputValue() as string;
```
daveadams56, line 21: I think the **previousValue** property is what we're using to render the existing value of the property returned by AM, where the is one. For example if we have a tree that is allowing the user to update their phone number, if they have provided it already and they already have a session (i.e. a "self service" tree). I think we should test with these kinds of trees as we move forward, to make this functionality easier to verify.

daveadams56 (commented):
1 of 2: I have reviewed most of the files in the PR, but will do the rest tomorrow as well as testing of the widget preview apps

src/lib/journey/journey.store.ts
```diff
@@ -84,52 +69,72 @@ export async function initialize(journey?: string): Promise<InitObject> {
...
-      submittingForm.set(false);
+      set({
+        completed: true,
+        error: null,
+        loading: false,
+        step: null,
+        successful: true,
+        response: nextStep.payload,
+      });
     } else if (nextStep.type === StepType.LoginFailure) {
```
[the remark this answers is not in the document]
daveadams56: On login failure it does not look like we "reset" the store as we do for login success or next step; Silly question but is this intentional?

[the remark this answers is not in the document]
daveadams56: Ok I think this is intentional since we want to capture the failure state for debugging purposes, however nothing seems to call reset() from within this file. Again I'm probably being silly, but want to check that we haven't forgotten to reset the state if there is a failure then the user tries again.

src/lib/components/compositions/checkbox/animated.svelte
cerebrl: Good point. I'm not sure if this will ever be used, but the `BooleanAttributeInputCallback`, which would use this checkbox component, can have a failed policies 🤷 

So, I have this available on this component, even though it may rarely, if ever be used.

src/lib/journey/journey.store.ts
```diff
@@ -139,32 +144,113 @@ export async function initialize(journey?: string): Promise<InitObject> {
...
+        }
+
+        /**
+         * If error code is 110, then the issue is just the authId expiring.
+         * If this is the first step in the journey, replace the callbacks
+         * with existing callbacks to resubmit with a fresh authId.
+         */
+        if (nextStep.payload?.detail && stepNumber === 1) {
+          const details = nextStep.payload?.detail as { errorCode: string };
+          if (details?.errorCode === '110') {
```
daveadams56, line 178: Might it be better to assign '110' to a descriptively named constant so it's easier to understand what it represents? I would just worry about your comment maybe getting re-written in the future and that context lost.

src/lib/journey/journey.store.ts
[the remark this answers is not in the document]
daveadams56: Ok, forget what I'm saying - I can see further down in **_journey.svelte_** that **reset()** is being called. Ignore me!

src/lib/journey/journey.svelte
```diff
@@ -1,97 +1,41 @@
...
-      let user = {
-        email: $email,
-        fullName: $fullName,
-        isAuthenticated: $isAuthenticated,
-      };
-
-      closeModal && closeModal();
-      returnUser && returnUser(user);
-      widgetDispatch && widgetDispatch('journey-success', user);
-    }
```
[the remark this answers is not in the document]
daveadams56: This is a lot nicer! I think it's the cleanest definition of a journey component I've seen, as they usually tend to be a bit messy - lots of different things going on in there. But this makes really clear the responsibility of this component. Nice!

src/lib/journey/stages/username-password.svelte
```diff
@@ -26,22 +55,26 @@
...
     </Button>
-  </Form>
-  <p class="tw_text-center tw_py-4 tw_text-secondary-dark dark:tw_text-secondary-light">
-    <T key="dontHaveAnAccount" html={true} />
-  </p>
-{:else if step.type === 'LoginSuccess'}
-  <T key="loginSucess" />
-{/if}
+    <p class="tw_text-center tw_py-4 tw_text-secondary-dark dark:tw_text-secondary-light">
+      <T key="dontHaveAnAccount" html={true} />
```
daveadams56, line 75: Just a thought about this link.. It seems to make sense if you are using the widget inline, redirecting to another version of the inline widget but presenting a registration journey. However when using the modal, having the link take you to another page which may or may not have the widget in a modal makes less sense. Perhaps this should switch the widget to the registration journey if it is in a modal, instead of redirecting?

src/widget/modal.svelte
```diff
@@ -0,0 +1,277 @@
+<script context="module" lang="ts">
```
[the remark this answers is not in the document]
daveadams56: I can see your earlier comment on inline.svelte now - I understand that this will be revisited in the future to reduce duplication, so this is not an issue

daveadams56 (commented):
Spent a bit more time looking at the larger changes and left a few comments, none of which are blocking or particularly high priority - just feedback for your awareness. I will find time tomorrow to focus on Chromatic and Storybook changes.

src/lib/journey/callbacks/kba/kba-create.svelte
cerebrl: Good catch, this is no longer needed. I will remove with my update later today.

src/lib/journey/callbacks/string-attribute/string-attribute-input.svelte
cerebrl: Yeah, the register journey is a perfect example of this. If you try to register, but your submission has some errors in it, the `input` property will have the previous value in it.

src/lib/journey/callbacks/terms-and-conditions/terms-conditions.svelte
cerebrl: Story: https://bugster.forgerock.org/jira/browse/SDKS-2004

src/lib/journey/journey.store.ts
cerebrl: This and the other caught by Dave will be fixed in my update.

src/lib/journey/journey.store.ts
cerebrl: This block of code has been removed and will be included in the update.

src/widget/modal.svelte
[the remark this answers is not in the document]
cerebrl: Yeah, this is related to this comment: https://github.com/cerebrl/forgerock-web-login-framework/pull/9/files#r934712797.

I recognize the duplication of code, but pulling it out into a reusable utility will take a while. There's a lot of closures that are referenced here and preserving all of this in a reusable way will take a bit. What's more, I didn't want to delay the testing of a lot of these assumption as well.

src/lib/journey/journey.store.ts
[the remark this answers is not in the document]
cerebrl: Just to make sure we are on the same page ... When we encounter a failure, we save off the failure message from the original request, and re-request the journey with no previous step to get a fresh set of callbacks. We then [`set` the new step with the old form failure message](https://github.com/cerebrl/forgerock-web-login-framework/pull/9/files#diff-e01603033c52bf6a043d215ff082252220802c8f537131e59b6fc3eb8de26cd3R223) on the store in order to communicate this failure to the user without presenting them a deadend. This is the main way of "resetting" the store for failures.

Now, the is a way to reset the store manually, and that's best for unrecoverable failures. This is the `reset` you may [see in the `journey.svelte` file](https://github.com/cerebrl/forgerock-web-login-framework/pull/9/files#diff-38f4b155f2dae7068eb3358d8b505a4b5730a0d12bc21bbec14ab47f89ba6bceR16). That happens when we failed on the first attempt and then something went wrong on the second attempt as well. To prevent looping, we bail out and show a UI that requires the user to restart the flow from scratch.

src/lib/journey/journey.store.ts
cerebrl: Good point. I will do that now.

src/lib/journey/journey.svelte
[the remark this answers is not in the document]
cerebrl: Thanks! :D

src/lib/journey/stages/username-password.svelte
cerebrl: This is all spot on. Currently, the links presented here don't do anything. There's a story that will address this, and it's found here: https://bugster.forgerock.org/jira/browse/SDKS-2053. We will end up with 4 distinct flows that will be internally managed by the widget:

1. Login
2. Registration
3. Forgot Password
4. Forgot Username

I just haven't accounted for the UX around these flows yet.

cerebrl:
@ryanbas21 

> I notice the login modal is loaded on page load rather than on button press, this could improve the TTI of the base page.

I'm doing this purely for the UX of opening the modal and the form is already rendered, but this is a choice left up to the implementer. You can see here I'm calling it within the `onMount`: https://github.com/cerebrl/forgerock-web-login-framework/pull/9/files#diff-7613af87cd0f9a8ee42f5fb8028bb00eb19577d1a51cd51291435a93956d9991R19. But, if it's not initialized by the time the modal is open, it will initialize itself. You can see that here: https://github.com/cerebrl/forgerock-web-login-framework/pull/9/files#diff-7a59713e60746c7bb5a10991dc81933bee66159a71bc8bd75c9ab5066979e548R141.

It's also worth noting that this initialization should not block the UI from rendering or from being interactive. If you "throttle" your Chrome connection to 3G, you can observe the fact that you can render the page and open the model prior to the `/authenticate`. Here's a GIF recording: https://share.getcloudapp.com/4gurq9dD. So, this pattern should be okay to demonstrate, but it should be completely optional.

...

After thinking about your comment, I'm wondering if you're asking about the import and mounting of the actual modal component itself, rather than the loading of the form. If you're suggesting we should asynchronously import the modal component on button click, rather than statically import the component, it's possible, but I'm not sure it's recommended for the e2e app as I like to keep them as fast and stable as possible (but, I get your point).

> We also load the error bundle as part of the modal, even if there is no error, I think this could also improve page scores if we only load JS that is necessary.

The widget component only has three files for it to functionally operate:

1. inline.js
2. modal.js
3. [variable-name]-[hash].js

(1) and (2) are the entry files for the form factors and (3) is the shared code between the 2 entry files, so that's were the bulk of the code is. So, only one of the entry files and the shared file would be loaded at once.

With that said, the "error bundle" as you call it is likely coming from SvelteKit itself or are the virtual files loaded by Chrome for source maps. But, the widget has no dedicated file or component for errors, so I doubt it comes from the widget.

> I know this is an "e2e" app, but this is how I would expect/instruct/hope someone uses a modal (loading it when required not always) so we should probably test it as such.

Oh, oops, I think this reinforces the fact that I misread your initial comment (I'm adding a second portion to the above ^^ ). I'll have to think about this a bit more.

> I'd like to highlight the great caching work being done (by svelte under the hood).

:)

> Do you know why we load a filename.svelte and a filename bundle? I couldn't figure out exactly the difference other than i think the svelte is a component bundle?

I'm not sure, tbh. This is SvelteKit being SvelteKit. As mentioned above, there are only the three files from the widget itself (only two of which would be loaded at one time). I'm currently not doing any code splitting, async imports or further refinement of the codebase. This would have to be a future story and prioritized by the team.

src/lib/journey/journey.store.ts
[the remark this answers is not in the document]
daveadams56: Ok this makes more sense now - so we want to present feedback from the previous failure whilst presenting a new attempt to authenticate. Got it. Thanks

src/lib/journey/callbacks/string-attribute/string-attribute-input.svelte
daveadams56: Ah ok so that's the same thing then, kind of. Understood

src/lib/journey/stages/username-password.svelte
daveadams56: Excellent, thanks

daveadams56 (approved):
This is shaping up really nicely! Thanks for your replies to my comments. None of the comments I had are blocking and you've addressed them anyway. Awesome work!

PR #13, opened 2022-09-07
feat(registration): complete registration flow and validation

cerebrl:
Another big PR as this has touched a lot of files. For that, I'm sorry. the large majority of the files that have been modified are fixing errors or warnings produced by our code quality tools.

Jira ticket: https://bugster.forgerock.org/jira/browse/SDKS-1815

## Summary

The registration flow has existed for a while. This PR improves on the UX of the flow and policy and validation reporting on a field level. 

## Features

- Implement input _info_ messaging as well as error
- Implement input list messaging for policies and policy failures
- Implements better control over simple versus complex messaging
- Implements finer grain control over policy logic and parsing
- Improves Storybook usage with interactions and newly added assertions
- Improves widget e2e testing to allow the passing of journeys through URL parameters

## Fixes

- Enables loading indicator on form submission button
- Adds custom security question option to KBA callback handler
- Adds better handling of the DS validation failure when IDM fails to report issue

## Updates

- Updated SvelteKit and Vite to latest


.github/workflows/ci.yml
```diff
@@ -26,10 +26,10 @@ jobs:
         run: npm run build
 
       - name: Test
-        run: npm run test
+        run: npm run ci
```
cerebrl, line 29: We now have `ci` and `test`. The latter is for running locally as I just wanted a quick to run tests without having to remember to run the build before.

README.md
```diff
@@ -52,7 +52,7 @@ Note: This project is currently in Alpha, so this is not available via public np
 
 1. `git clone https://github.com/cerebrl/forgerock-web-login-framework`
 2. `cd forgerock-web-login-framework`
-3. `npm run install`
```
cerebrl, line 55: I never realized this no longer worked. Not sure when that happened, but 🤷

src/lib/components/_utilities/locale-strings.svelte
```diff
@@ -1,5 +1,5 @@
 <script lang="ts">
-  import { interpolate } from '$lib/utilities/i18n.utilities';
+  import { interpolate } from '$lib/_utilities/i18n.utilities';
```
cerebrl, line 2: You're going to see quite a bit of these path and directory changes. I wanted to create a consistent and obvious way to declare utilities for a group of components. This helps bring that utilities directory to the top of the file list, and makes it obvious that it stands apart from the regular components.

src/lib/components/compositions/checkbox/animated.stories.js
```diff
@@ -1,5 +1,5 @@
 import { expect } from '@storybook/jest';
-import { screen, userEvent } from '@storybook/testing-library';
+import { userEvent, within } from '@storybook/testing-library';
```
cerebrl, line 2: These changes were recommended by Storybook, likely for the upcoming v7.

src/lib/components/compositions/checkbox/animated.story.svelte
```diff
@@ -6,32 +6,34 @@
   import Form from '$components/primitives/form/form.svelte';
 
   export let checkValidity: ((event: Event) => boolean) | null = null;
-  export let errorMessage = '';
+  export let message = '';
```
cerebrl, line 9: Changing from `errorMessage` to just generic `message` across all components as we might want to communicate non-error information to the user at the field level.

src/lib/components/compositions/checkbox/animated.svelte
```diff
@@ -1,16 +1,17 @@
...
 
   export let checkValidity: ((event: Event) => boolean) | null = null;
-  export let errorMessage = '';
+  export let message = '';
   export let firstInvalidInput: boolean;
   export let isRequired = false;
   export let isInvalid: boolean | null = null;
   export let key: string;
   export let onChange: (event: Event) => void;
+  export let showMessage: boolean | undefined = undefined;
```
cerebrl, line 14: Adding an explicit control over showing the message or not. I could no longer determine whether to show it implicitly.

src/lib/components/primitives/form/form.svelte
```diff
@@ -55,6 +55,11 @@
           let messageKey = input.getAttribute('data-message');
           input.setAttribute('aria-describedby', messageKey || '');
 
+          if (messageKey) {
+            const messageEl = document.getElementById(messageKey)
+            messageEl?.setAttribute('data-error', 'true');
+          }
```
cerebrl, line 61: I could set a class on the message element, rather than a data attribute, but I'm on the fence.

src/lib/components/primitives/message/input-message.stories.js
```diff
@@ -0,0 +1,25 @@
...
+    key: 'infoMessage',
+    type: 'info'
+  },
+};
+
+export const Error = {
+  args: {
+    message: 'Please review your input.',
+    key: 'infoMessage',
+    type: 'info'
```
cerebrl, line 23: Darn it! This should be type "error". I'll fix this along with any feedback I get from you all.

src/lib/components/primitives/message/input-message.svelte
```diff
@@ -0,0 +1,30 @@
+<script lang="ts">
```
cerebrl, line 1: New generic input message component. It replaces the old error message component.

src/lib/components/primitives/select/select.svelte
```diff
@@ -13,15 +13,34 @@
...
 
   let inputEl: HTMLSelectElement;
+  let shouldDisplayOption = true;
+
+  /**
+   * If label and option share the same text, only display option
+  */
+  if (defaultOption === null && (options[0].text === label)) {
+    shouldDisplayOption = false;
+  }
```
cerebrl, line 26: The comment mostly says it, but it's just visual. This prevents redundant label and option messaging.

src/lib/journey/_utilities/step.utilities.ts
```diff
@@ -0,0 +1,70 @@
+import {
```
cerebrl, line 1: A lot of this file contains functions that have been moved out of the `journey.store.ts` file, so the logic can be tested independently. Much of this is for determining when and how to handle the varying levels of errors in a journey. Once such error is the DS error when IDM fails to validate the callbacks.

src/lib/journey/callbacks/_utilities/callback.utilities.ts
```diff
@@ -0,0 +1,374 @@
...
+        console.log(`Parsing failure for ${label}`);
+      }
+    } else {
+      return policy as PolicyRequirement;
+    }
+  });
+}
+
+/** *********************************************
+ * OLD METHODS
```
cerebrl, line 304: The below are old methods that we need to remove once we can validate we don't need them anymore. The "new" methods from above should cover all use cases we need.

src/lib/journey/callbacks/_utilities/policies.svelte
```diff
@@ -0,0 +1,62 @@
+<script lang="ts">
```
cerebrl, line 1: This is the new policies component that will display an input's policies for validating the value. This way it's reusable between all the callbacks that use policies.

src/lib/journey/callbacks/kba/kba-create.svelte
```diff
@@ -18,18 +18,47 @@
...
   let value = '';
 
-  callback.setQuestion(questions[0].text);
+  /**
+   * `getOutputValue` throws if it doesn't find this property. There _may_ be a context
+   * in which the property doesn't exist, so I'm going to wrap it in a try-catch, just
+   * in case
+  */
+  try {
+    shouldAllowCustomQuestion = callback.getOutputValue('allowUserDefinedQuestions') as boolean;
```
cerebrl, line 43: The KBA callback has an option to allow custom questions. So, most of this new code is to allow this feature. 

Being that I'm using this low level callback feature (we really should add a higher level API for this in the SDK's KbaCallback handler), it throws when this `allowUserDefinedQuestions` doesn't exist, so in order to keep the widget from exploding I'm wrapping it in a try-catch.

cerebrl: Oh, I'm not sure if it's possible for the KBA callback to NOT have the `allowUserDefinedQuestions` property, but I have quite a few example KBA callbacks around that did indeed NOT have it, so that's why I'm adding the try-catch, just in case.

src/lib/journey/callbacks/password/validated-create-password.svelte
```diff
@@ -1,32 +1,48 @@
...
+<Base
+  {callback}
+  {firstInvalidInput}
+  {idx}
+  {isInvalid}
+  {isRequired}
+  key={inputName}
+  showMessage={!validationFailures.length}
+>
+  <Policies {callback} key={inputName} label={prompt} messageKey="passwordRequirements" />
```
cerebrl, line 47: Here's the usage of the new Policies component. Notice how we deactivate the simple message (line above) that comes with the Base component if there are validation failures and use the more complex Policies component.

src/lib/journey/callbacks/string-attribute/string-attribute-input.svelte
```diff
@@ -1,26 +1,31 @@
...
+  } from '$journey/callbacks/_utilities/callback.utilities';
   import type { AttributeInputCallback } from '@forgerock/javascript-sdk';
 
+  import {
+    getValidationFailures,
+  } from '$journey/callbacks/_utilities/callback.utilities';
   import Input from '$components/compositions/input-floating/floating-label.svelte';
-  import { interpolate } from '$lib/utilities/i18n.utilities';
+  import { interpolate } from '$lib/_utilities/i18n.utilities';
+  import Policies from '$journey/callbacks/_utilities/policies.svelte';
```
cerebrl, line 13: Similar changes to this and the `ValidateCreateUsername` callback handler since they all have policies.

src/lib/journey/journey.store.ts
```diff
@@ -140,43 +132,51 @@ export function initialize(initOptions?: StepOptions): JourneyStore {
...
+         *
+         * Assert that restartedStep is FRStep as that is required for the above condition to be true.
+         * Also, assert that previousCallbacks is FRCallback[] as that too is required for above to be true.
+         *
+         * Attempt a refactor using Ryan's suggestion found here: https://www.typescriptlang.org/play?#code/PTAEHUFMBsGMHsC2lQBd5oBYoCoE8AHSAZVgCcBLA1UABWgEM8BzM+AVwDsATAGiwoBnUENANQAd0gAjQRVSQAUCEmYKsTKGYUAbpGF4OY0BoadYKdJMoL+gzAzIoz3UNEiPOofEVKVqAHSKymAAmkYI7NCuqGqcANag8ABmIjQUXrFOKBJMggBcISGgoAC0oACCbvCwDKgU8JkY7p7ehCTkVDQS2E6gnPCxGcwmZqDSTgzxxWWVoASMFmgYkAAeRJTInN3ymj4d-jSCeNsMq-wuoPaOltigAKoASgAywhK7SbGQZIIz5VWCFzSeCrZagNYbChbHaxUDcCjJZLfSDbExIAgUdxkUBIursJzCFJtXydajBZJcWD1RqgJyofGcABqDGg7EgAB4cAA+AAUq3y3nBqwUPGEglQlE4IwA-FcJcNQALOOxENJvgBKUAAb0UJT1CNAPNQ7SJoIAvBbQAAiZWq75WzV0hmgUG6vXg6CCFBOsheVZukoB0CKAC+incNCGUtAZtpkHpvuZrI54slzF5VoAjA6ANzkynUrxCYjyqV8gWphUAH36KrVZHVAuB8BaXh17oNRpNqXNloA5JWpX3Ne33XqfZkyGy8+6w0GJziWV683PO8XS8wjXFmOqR0Go8wAhlYKzuPoeVbsNBoPBc6HgiocM0PL7QIh4H0GMD2JG7owpewDDMJA-AnuoiRfvAegiF4VoAKKrAwiALPoVpJNiVrgA4qADqAABykASFaQQqAA8l8ZDvF6-DAUcqCOAorjSHgcbvjoCpfF6aKINCwiXF8kgftEIgGBw2ILEwrAcDwQQlEAA
```
cerebrl, line 155: Didn't have time to refactor this TypeScript issue we discussed, but I left this comment, so we can easily revisit it.

src/lib/journey/stages/registration.svelte
```diff
@@ -71,7 +72,7 @@
         {firstInvalidInput}
       />
     {/each}
-    <Button width="full" style="primary" type="submit">
+    <Button busy={loading} style="primary" type="submit" width="full">
```
cerebrl, line 75: We might want to deactivate the button while it's loading, but we'd probably also want to add a timer or something ... not sure. I really don't like deactivating buttons if I don't have to.

src/locales/us/en/index.json
```diff
@@ -3,8 +3,15 @@
   "dontHaveAnAccount": "Don’t have an account? <a href='/register'>Sign up here!</a>",
   "closeModal": "Close Modal",
   "chooseDifferentUsername": "Please choose a different username.",
-  "ensurePasswordIsMoreThan": "Ensure password contains more than {minPasswordLength} characters.",
+  "constraintViolationForPassword": "Your password does not meet the acceptance criteria.",
```
cerebrl, line 6: Lot's of strings!!!! I need to go through all of these and clean them up at some time.

.svelte-kit/tsconfig.json
```diff
@@ -27,12 +27,16 @@
 	},
 	"include": [
 		"ambient.d.ts",
+		"../vite.config.ts",
```
ryanbas21, line 30: I wonder if we can just remove the `includes` property as includes should include everything not in `exclude`. maybe theres more of a reason for include than I realize though.

src/lib/components/primitives/message/input-message.svelte
```diff
@@ -0,0 +1,29 @@
...
+  import T from '$components/_utilities/locale-strings.svelte';
+
+  export let message: string;
+  export let key: string | undefined = undefined;
```
ryanbas21, line 5: non-blocking, but i spent some time trying to find out if there is an equivalent to a `maybe` type in vanilla typescript. there isn't, of course, but we could do something like this:

```suggestion
 type Maybe<T> = T | null | undefined;
 export let key: Maybe<String> = undefined;
```

I'd probably abstract that type out of the file though to reuse it. Is there a signficant value increase in this? probably not outside of if we debate readability of it. So i'll leave this here as a note and see if you like it more or not.

.svelte-kit/tsconfig.json
cerebrl: This is a file controlled by SvelteKit. If we made any changes here, I'm sure they'd get overwritten.

src/lib/components/primitives/message/input-message.svelte
cerebrl: I actually like this idea, I can look at updating some of instances of these, but I'd recommend slowly making this change through the whole project as there are likely a lot of instances of this.

ryanbas21 (approved):

rodrigoareis (approved):
Changes looks good to me

PR #17, opened 2022-09-20
feat: text-output-support

ryanbas21:
Adds basic support for text-output and suspend-text-output nodes

https://bugster.forgerock.org/jira/browse/SDKS-2006


src/lib/journey/callbacks/suspend-text-output/suspend-text-output.svelte
```diff
@@ -0,0 +1,13 @@
+<script lang="ts">
+  import type { SuspendedTextOutputCallback } from '@forgerock/javascript-sdk';
+  import Centered from '$components/primitives/box/centered.svelte';
```
cerebrl, line 3: I try to separate the external dependency imports from the internal components with an empty line.

src/lib/journey/callbacks/text-output/text-output.stories.js
```diff
@@ -0,0 +1,43 @@
...
+
+  await userEvent.tab();
+
+  await userEvent.click(yes);
+
+  const no = screen.getByText('No');
+  await userEvent.click(no);
+
+  await userEvent.tab();
+  }
```
cerebrl, line 41: There's some indentation issue here.

cerebrl: Also, since this an interactive component, let's keep the interaction portion, but I'd like to ask it to be consistent with the other implementations. The usual pattern is the following:

1. Base: just the simple, default static rendering of the component. No "play" prop here
2. <Additional variant> (optional): static variants of the component, like primary, secondary, if applicable
3. Error: a static, forced error state of the component
4. Interaction: This is the one where we "play" the interaction and make assertions against the result

What's more, we create the automated "interaction"/"play" version through composition with one of the previous static variants.

```js
const Template = (args) => ({
  Component: Base,
  props: args,
});

export const Interaction = Template.bind({});

Interaction.args = { ...DialogWithTrigger.args };

Interaction.play = async ({ canvasElement }) => {
    // interaction code
}
```

This pattern allows for more robust compositions, and is the one found in most Storybook docs as far as I've seen.

Finally, if we have an interaction, I'd actually like to take the opportunity to assert things against it. We can pull in `expect` from the `jest` library. You can see a simple, yet thorough, example here: https://github.com/cerebrl/forgerock-web-login-framework/blob/main/src/lib/components/compositions/checkbox/animated.stories.js

src/lib/journey/callbacks/text-output/text-output.svelte
```diff
@@ -0,0 +1,31 @@
...
+<Centered>
+  <div>{message}</div>
+  {#each options as opt, index}
+    <Button
+      style={selected === index ? 'primary' : 'secondary'}
+      type="button"
+      width="auto"
+      onClick={() => setValue(index)}
+    >
+      {opt}
```
cerebrl, line 28: Are these options localizable within ForgeRock? In other words, does the admin set the option in English, Spanish, French ... Or, is it just one value per option? If it's the former, we can just render it how you're doing it here.

But, if it's the latter (not localization in ForgeRock), then we need to construct a key out of the option, and use that as a lookup to a content string from the locale file.

cerebrl (changes requested):
Thanks for the PR. Left a bunch of comments to help with conventions, patterns and such. We need to think about two things throughout this new code addition:

1. The type of string in the message (plaintext, HTML or script)
2. Localization

Finally, we should also add an e2e test for this new component. You should have admin access to my tenant, so you can add a special tree for such testing purposes.

Let me know if you have any questions.

cerebrl:
@ryanbas21 Also, the Chromatic publishing seems to have an error regarding missing project token.

ryanbas21:
> @ryanbas21 Also, the Chromatic publishing seems to have an error regarding missing project token.

Thats because this is your repo. I don't have access to your secrets.

cerebrl:
> Thats because this is your repo. I don't have access to your secrets.

The project token is hardcoded in the `package.json`. It's definitely no secret right now :)

cerebrl:
Also, you should be able to just straight push to this repo as you're a collaborator on it. So, you shouldn't need to use your own fork.

src/lib/journey/callbacks/text-output/text-output.svelte
ryanbas21: Looks like it is the former.
Screen Shot 2022-09-26 at 1 51 25 PM

PR #18, opened 2022-09-23
feat(i18n): improve internationalization and refactor code

cerebrl:
Jira ticket: https://bugster.forgerock.org/jira/browse/SDKS-2032

## Summary

The original work to support internationalization had some flaws and weaknesses. This PR improves on the original implementation and addresses some comments from the prior PR. 

## Features

- Adds the proper locale content globally to Storybook
- Makes typing more strict to catch locale file property misalignment
- Add content key normalization much more robust for transforming various strings into camelCase
- Improves overall typing throughout the project
- Removes unneeded console logs
- Add localization support to the ChoiceCallback

## Out of scope

- Translation of content strings in other languages; there's currently English intermixed with other languages in the language files


.storybook/preview.js
```diff
@@ -1,7 +1,12 @@
 import { themes } from '@storybook/theming';
 
+import { initialize } from '../src/lib/locale.store';
+
 import '../src/app.css';
 
+// Initializes localized content globally for all stories
+initialize();
```
cerebrl, line 8: This adds the true, fallback locale strings globally to Storybook. This means the components won't fallback to converting the key itself to the displayed string.

src/lib/_utilities/i18n.utilities.test.ts
```diff
@@ -79,3 +79,113 @@ describe('Test interpolate utility function', () => {
     expect(string).toBe(expected);
   });
 });
+
+describe('Test text-to-key function', () => {
```
cerebrl, line 83: The function tested here is the `textToKey` which allows us to use any unique string as a key for content lookup. It essentially converts anything thrown at it to camelCase to normalize all the keys in the locale files. 

I added as many different variations to strings I could think of to test. If you can think of others, just let me know.

src/lib/_utilities/i18n.utilities.ts
```diff
@@ -109,7 +109,48 @@ export function interpolate(
     messageDirty = externalText;
   }
 
-  const messageClean = xss(messageDirty);
+  const messageClean = sanitize(messageDirty);
 
   return messageClean;
 }
+
+export function textToKey(text: string | number) {
```
cerebrl, line 117: This was a wild ride of regex experimentation :)

src/lib/journey/callbacks/unknown/unknown.svelte
```diff
@@ -0,0 +1,9 @@
+<script lang="ts">
+  import type { FRCallback } from '@forgerock/javascript-sdk';
+
+  export let callback: FRCallback;
+
+  const type = callback.getType();
+</script>
+
+<p>Unknown callback encountered: {type}. Please contact support.</p>
```
cerebrl, line 9: We probably want to do something better than this, but I needed a fallback component for the `mapCallbackToComponent` function.

src/lib/journey/journey.interfaces.ts
```diff
@@ -2,20 +2,52 @@ import type { FRStep, FRLoginFailure, FRLoginSuccess } from '@forgerock/javascri
...
+}
+export type StepTypes = WidgetStep | FRLoginSuccess | FRLoginFailure | null;
+export interface WidgetStep extends FRStep {
+  /**
+   * TODO: It's complicated, but this `any` is needed for `callback` compatibility with
+   * the `mapCallbackToComponent`. FRStep types this callback array as `FRCallback[]`
+   * which is not _exactly_ correct as it's an array of various types of callbacks, all
+   * which extend FRCallback. So, there are heterogeneous types within the array.
+   */
+  callbacks: any[];
```
cerebrl, line 52: To get a perfectly clean `npm run check`, I had to override the `callbacks` property on `FRStep` type with an `any[]`. the original type is `FRCallback[]`, but there's not enough overlap between all the actual callbacks, like `ChoiceCallback`, `NameCallback`, `KBACallback` for the callback components.

The issue is at the exact place the type error occurs, you can't type the value since it's in the Svelte template portion of the code :(

src/lib/locale.store.ts
```diff
@@ -1,42 +1,68 @@
...
+    tryAgain: z.string(),
+    useValidEmail: z.string(),
+    unrecoverableError: z.string(),
+    unknownNetworkError: z.string(),
+    usernameRequirements: z.string(),
+    validatedCreatePasswordCallback: z.string(),
+    validatedCreateUsernameCallback: z.string(),
+    valueRequirements: z.string(),
+  })
+  .strict();
```
cerebrl, line 64: Adding this `strict` property, it will report unknown properties as well, not just ignore them. This is good for accidentally misspelling a prop.

src/locales/ca/en/index.json
```diff
@@ -3,8 +3,15 @@
   "dontHaveAnAccount": "Don’t have an account? <a href='/register'>Sign up here!</a>",
   "closeModal": "Close Modal",
   "chooseDifferentUsername": "Please choose a different username.",
-  "ensurePasswordIsMoreThan": "Ensure password contains more than {minPasswordLength} characters.",
+  "constraintViolationForPassword": "Your password does not meet the acceptance criteria.",
```
cerebrl, line 6: It's worth noting that the only locale file the Widget actually uses is the `us/en/index.json`. The Widget expects any locale override to be provided through the Widget instantiation. These other content files are not actually used by the Widget itself, but used for testing and the future app.

src/routes/(app)/+page.server.ts
```diff
@@ -1,28 +1,23 @@
...
+  let localeContent: { default: z.infer<typeof stringsSchema> };
 
-  // TODO: Reworking typing here to index the locales object
-  const locale = getLocale(userLocale, '_') as 'ca_en' | 'ca_fr' | 'us_en' | 'us_es';
+  try {
+    localeContent = await import(`../../locales/${country}/${lang}/index.json`);
+  } catch (err) {
+    console.error(`User locale content for ${userLocale} was not found.`);
+    localeContent = await import(`../../locales/us/en/index.json`);
+  }
```
cerebrl, line 20: Finally figured out how to do async imports properly \o/

src/routes/e2e/widget/inline-instance/+page.svelte
```diff
@@ -11,6 +11,7 @@
 
   let formEl: HTMLDivElement;
   let userInfo: Record<string, string> | null;
+  // TODO: Use a more specific type
   let widget: any;
```
cerebrl, line 15: These are all the e2e test routes, so I'm not all that worried about the typing here right now.

tsconfig.json
```diff
@@ -1,7 +1,6 @@
 {
   "extends": "./.svelte-kit/tsconfig.json",
   "compilerOptions": {
-    // @rollup/plugin-typescript TS1259: Module '"/Users/justin.lowery/Github/forgerock-web-login-framework/src/locales/ca/en/index"' can only be default-imported using the 'allowSyntheticDefaultImports' flag
```
cerebrl, line 4: This comment was not perfectly accurate, but I made a second attempt at importing JSON into the project without the synthetic import property. Unfortunately, there were too many different tools that I could not get to align when this was removed.

cerebrl (commented):
Making my explanatory comments.

src/lib/_utilities/i18n.utilities.test.ts
ryanbas21: wow you got a lot of cases covered here!

src/lib/journey/journey.interfaces.ts
ryanbas21, line 52: There may be something we can do with our new found knowledge however i havent thought of anything just yet.

ryanbas21 (changes requested):
Initial comments as I think more on other things

src/lib/_utilities/i18n.utilities.ts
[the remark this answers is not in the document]
cerebrl: We could use Zod for this, but it feels like overkill. My main intention with it was to help enforce developer facing interfaces that are fed into the program by enforcing correct types both in build time and run time. I'm not using it much internally.

One thing to note is Zod's `parse` will `throw` if it fails. When we handle internal functionality like this, _if_ we did use Zod, I'd prefer `safeParse`, so it doesn't `throw`. With `safeParse`, we'd have to use a conditional to check for errors from the `safeParse`. So, with all things considered, it would likely complicate the function, rather than simplify.

src/lib/journey/journey.interfaces.ts
cerebrl: This is pretty isolated and has no real affect on the rest of the system, so I'm not too concerned with it. It's more me trying to wrangle the perfectionist in me :)

src/lib/journey/_utilities/map-callback.utilities.ts
```diff
@@ -10,13 +10,14 @@ import Name from '$journey/callbacks/username/name.svelte';
...
 import TermsConditions from '$journey/callbacks/terms-and-conditions/terms-conditions.svelte';
+import Unknown from '$journey/callbacks/unknown/unknown.svelte';
 
 /**
  * Iterate through callbacks received from AM and map the callback to the
  * appropriate callback component, pushing that component
  * the StepComponent's array.
  */
-export function mapCallbackToComponent(cb: any) {
+export function mapCallbackToComponent(cb: FRCallback) {
```
daveadams56, line 20: I read your comment on [[src/lib/_utilities/i18n.utilities.ts](https://github.com/cerebrl/forgerock-web-login-framework/pull/18/files/0ceb5f624300f6f2b82824def819da53f86668c3#diff-d521444e98934562466afd9f553b7b765b61b1d56f36cf33cdf5e8b69739213b)](https://github.com/cerebrl/forgerock-web-login-framework/pull/18/files/0ceb5f624300f6f2b82824def819da53f86668c3#diff-9a05e063ad2fef3b804a33373dbd21f71a799916be581e8289aee521b218bf7d) about there not being enough overlap between callbacks to actually use FRCallback, and instead you used `any`. I can see here you've swapped `any` for `FRCallback` and I am wondering do we have an issue with the types in the SDK, if not all of them are comparable / are subclasses of the FRCallback class? Maybe I'm a bit confused but it would be good to know the issue there

src/lib/journey/_utilities/map-stage.utilities.ts
```diff
@@ -2,14 +2,19 @@ import Generic from '$journey/stages/generic.svelte';
...
+export function mapStepToStage(currentStep: StepTypes) {
   /** *********************************************************************
    * SDK INTEGRATION POINT
    * Summary:SDK step method for getting the stage value
    * ----------------------------------------------------------------------
    * Details: This method is helpful in quickly identifying the stage
    * when you want to provide special layout or handling of the form
    ********************************************************************* */
+  if (!currentStep || currentStep.type !== 'Step') {
+    return Generic;
```
daveadams56, line 16: What is the meaning of Generic in this instance? I'm a little confused as to what this could mean in a Step

src/lib/journey/callbacks/choice/choice.svelte
```diff
@@ -14,13 +15,30 @@
...
-    ?.map((label, idx) => ({ text: label, value: `${idx}` }));
+  const choiceOptions = callback.getChoices()?.map((text, idx) => ({
+    /**
+     * Since locale content keys for the choice component are built off of the
+     * values, there will not be any existing key-value pairs in the provided
+     * content. The third argument here, the original value, is what will be
+     * displayed. If you want to localize it, you'll need to add content keys
+     * in the locale file for that to override the original value.
+     */
+    text: interpolate(textToKey(text), null, text),
```
daveadams56, line 26: Just want to make sure I understand this properly - we are providing the ability for a developer to define a set of choices in a choice node on the server, and localize this by providing matching values in a translation file within the app? Is that what's happening here?

src/lib/journey/callbacks/unknown/unknown.svelte
daveadams56: Thinking about it, perhaps we might want to add something to the journey object that we expose, to show that an unknown callback was encountered?

src/lib/journey/journey.store.ts
```diff
@@ -37,10 +37,8 @@ export function initialize(initOptions?: StepOptions): JourneyStore {
      * form failure due to 400 response from ForgeRock.
      */
     let previousCallbacks: FRCallback[] | undefined;
-    let previousStage: string | undefined;
 
     if (prevStep && prevStep.type === StepType.Step) {
-      previousStage = prevStep?.getStage && prevStep.getStage();
```
daveadams56, line 43: I don't have a problem with it but am a little curious - why are we removing the reference to `previousStage` and what were we doing with it before this PR?

src/lib/locale.store.ts
daveadams56: Maybe I misunderstand what we are doing here (see my other comment on Choice Callback), but do we not want to allow additional properties, in case they occur in a node on the server (which is detached from this)?

src/lib/user/user.store.ts
```diff
@@ -1,16 +1,18 @@
...
   get: (getOptions?: ConfigOptions) => void;
   reset: () => void;
 }
 export interface UserStoreValue {
   completed: boolean;
-  error: {
-    code?: number | null;
-    message: string | null;
-  } | null;
+  error: Maybe<{
```
daveadams56, line 12: I'm seeing `Maybe` pop up quite often in this PR - I had to look up what the Maybe pattern is, and now I think I understand. Is there any tangible difference to the way we handle nulls etc by using this pattern, or is it just for reducing duplication and increasing consistency within the app?

src/routes/(app)/+page.server.ts
daveadams56: Nice! I think this change also means that we are no longer hard coding locales, instead letting the naming of the files define the locales available - right?

daveadams56 (approved):
I think I understood what you were trying to do here and overall this looks good. None of my comments are blocking, they're more for me to understand how localization stuff works within the widget, and some of the patterns you're using. If you get time to respond to them, it would be interesting to hear your thoughts. I'm going to approve this and hopefully await your responses, thanks!

.storybook/preview.js
```diff
@@ -1,7 +1,12 @@
...
+
 // Set background color quicker to ensure Accessibility contrast is properly measured
 // const prefersDarkTheme = window.matchMedia('(prefers-color-scheme: dark)').matches;
```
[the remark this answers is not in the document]
cerebrl: Yeah, this is intentional. I'm not sure how we are going to implement dark mode into Storybook for the long term.

src/lib/_utilities/i18n.utilities.ts
cerebrl: I'm just changing the name of the import, so it makes more sense semantically. Even though the docs use `xss` as the imported function name, I never made sense to me, because it essentially says "cross-site script this value". I prefer "sanitize this value" :)

src/lib/journey/_utilities/map-callback.utilities.ts
cerebrl: Yeah, the way the SDK is typed, this gets a bit hairy when assignment is done in templates. Like the call site of this very function in the stage component. The issue is I can't assign types in the template portion of Svelte (the HTML essentially), so I have to make adjustments higher up in the code flow.

I'm not entirely sure what our long term plan is concerning this, but considering the downstream components are properly typed, I'm not super concerned about this. I doubt properly typing this would lead to an equivalent value add to the actual code itself.

Switching this from `any` to `FRCallback` didn't produce any side-effects, and it removes another `any` from the codebase. We now just have the one from the portion of code you linked to.

src/lib/journey/_utilities/map-stage.utilities.ts
cerebrl: You can think of `Generic` in this sense as an unnamed step. So, when there's no `stage` value provided in the step, this `Generic` component is used. When there is a `stage` value, it will use the the named stage component, if matched.

The reason this particular condition is used is prior to the `switch` below is if the app is still booting up, there is no "step" data, so we need to render the `Generic` component as the named stage components have certain assumptions about the step data.

src/lib/journey/callbacks/choice/choice.svelte
cerebrl: Yes, just to reiterate:

This framework prioritizes internal, localized content, when available. I'm doing this because having a single source of truth for your content is desirable. I want to minimize the amount of locale content that's spread between this and AM. Most companies have a separate team that manages the locale content, and they are used to dealing with the content within the projects themselves or locale (micro) services.

Now, there are some nodes in AM that neither provide a way to localize values NOR provide a desirable value to "key off of" for locale content. This is why I wrote the `textToKey` utility method. It provides a way to convert a regular string with spacing and punctuation into a normalized key. Platform UI does a same thing.

The choice node is one such example of a node that provides no way to localized the values and doesn't provide a natural key to use.

Now, the `interpolate` function works like this:

1. First argument is the key to look up the localized content string
2. Second is the data that is used for interpolation (injecting dynamic values into static strings)
3. Third is the fallback string, so if no content is found matching that key, it will use the text provided in this argument

There is also the chance when the key doesn't resolve to a localized string, and no fallback text is provided, it will convert the very key itself into a string. This is the fallback for the fallback :)

src/lib/journey/callbacks/unknown/unknown.svelte
cerebrl: By "journey" object, you mean the object provided to the application upon the `onSuccess` event? If so, that's in interesting idea, but I'd have to really think about how to do that. The way this code works is it's very unidirectional. Data flows from parent to child, but not child to parent. This is intentional and goes back to the Flux architectural pattern from Facebook's React team.

Being that this is deeper in the app than the store that is returned to the app, the communication would be against the natural flow of the code. But, that doesn't mean we couldn't figure out a way to do such a thing.

src/lib/journey/journey.store.ts
cerebrl: This was vestigial code that should have been removed a while ago, but it was left behind. I used to use the step's stage value to do a comparison for detecting if the user encountered some kind of validation issue, like in registration. But, a better method replaced it quite a while ago.

src/lib/locale.store.ts
cerebrl: Currently, this object exactly lines up with the locale files that we have in the project. Even if the developer wanted to add a property, it wouldn't result in any content lookup because the content file wouldn't have the key. It would then fallback to the provided text from AM.

This is a problem that we will really need to think about, but I'm unsure how to solve it without real world testing and feedback. We will likely iterate on how content is handled in this framework multiple times in the coming months.

src/lib/user/user.store.ts
cerebrl: More the latter. Just reduces the `thing | null | undefined` littered all over the place.

src/routes/(app)/+page.server.ts
cerebrl: That is correct.

PR #19, opened 2022-09-26
feat(Sdks/2006):add text output and suspend text output support

ryanbas21:
Jira: https://bugster.forgerock.org/jira/browse/SDKS-2006

## Summary

This PR adds the support for `SuspendedTextOutputCallback` and `TextOutputCallback` with a single component, as they are the same, essentially. There's also an addition, though not fully complete as it needs a deeper UX discussion, of the `ConfirmationCallback`.


src/lib/journey/callbacks/text-output/text-output.svelte
```diff
@@ -0,0 +1,35 @@
...
+  export let callback: TextOutputCallback;
+  export let choice: ConfirmationCallback;
+
+  const options = choice.getOptions();
+
+  $: message = callback.getMessage();
+
+  const sanitized = xss(message);
+
+  $: selected = choice.getDefaultOption(); // gets the index position
```
cerebrl, line 16: Have you pushed your most recent changes? This seems like the older, incorrect version of the component we discussed yesterday.

src/lib/journey/callbacks/suspend-text-output/suspend-text-output.svelte
```diff
@@ -0,0 +1,11 @@
+<script lang="ts">
```
cerebrl, line 1: There should already be a directory named `suspended-text-output`, can you use that directory and the same word, "suspended", for the filename. That way, it will match the callback name.

src/lib/journey/callbacks/suspend-text-output/suspend-text-output.mock.ts
```diff
@@ -0,0 +1,20 @@
...
+    "callbacks": [
+        {
+            "type": CallbackType.SuspendedTextOutputCallback,
+            "output": [
+                {
+                    "name": "message",
+                    "value": "An email has been sent to the address you entered. Click the link in that email to proceed."
```
cerebrl, line 11: Is it possible for this to contain HTML?

src/lib/journey/callbacks/text-output/text-output.mock.ts
```diff
@@ -11,3 +13,56 @@ export const docsExample = {
...
+
+export default {
+  authId: 'foo',
+  "callbacks": [
+    {
+      "type": "TextOutputCallback",
+      "output": [
+        {
+          "name": "message",
+          "value": "Had coffee?"
```
cerebrl, line 25: Can we have multiple variants of this callback? Some with plaintext, some with HTML, some that are malicious code?

cerebrl (changes requested):
Left some comments, but I'm assuming this is still a work-in-progress, yeah? I say that because it's missing the inclusion of this callback component with the rest of the callbacks so it can be used functionally within the app/widget.

daveadams56 (commented):
I have a couple of questions about whether we want to parameterise certain things in the tests, and one about the naming of the test output callback. 

> Update - Just noticed that the preview app doesn't work - looks like a CORS error?
> Screenshot 2022-10-04 at 16 46 32

Update - I was using the wrong preview app URL, so the CORS issue I previously mentioned is no longer an issue.

Should be happy to approve once I can get answers to these points and understand a little better if there are any more changes / fixes to come in this PR?

tests/widget/modal/widget-modal.login-with-confirm.test.js
```diff
@@ -0,0 +1,24 @@
+import { expect, test } from '@playwright/test';
+
+test('Modal widget with simple login and confirmation', async ({ page }) => {
+  await page.goto('widget/modal?journey=LoginWithConfirmation');
+
+  const loginButton = page.locator('button', { hasText: 'Open Login Modal' });
+  await loginButton.click();
+
+  await page.fill('text="User Name"', 'demouser');
```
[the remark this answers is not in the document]
cerebrl: Yes, there's a future story to improve CI coming up, so we can address these improvements then.

src/lib/journey/callbacks/text-output/text-output.mock.ts
```diff
@@ -1,5 +1,67 @@
-export const docsExample = {
-  type: 'TextOutputCallback',
+import { CallbackType } from '@forgerock/javascript-sdk';
+
+export default {
+  authId: 'foo',
+  callbacks: [
+    {
+      type: CallbackType.TextOutputCallback,
```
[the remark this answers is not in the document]
cerebrl: Yes, since `SuspendedTextOutputCallback` extends `TextOutputCalllback` with zero additional behavior, it's essentially a copy. The other variations of "suspend" were just a typo.

Rather than having identical components for the sake of matching callbacks with different names, I went with one callback to handle both. That way we don't have to keep two components in-sync if something changes.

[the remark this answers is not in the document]
daveadams56: Makes sense to me - just wanted to check what we decided as there was a conversation on this further up. Thanks

src/lib/journey/callbacks/confirmation/confirmation.svelte
```diff
@@ -0,0 +1,80 @@
...
+
+{#if displayType === 'select'}
+  <Select
+    {firstInvalidInput}
+    isRequired={false}
+    key={inputName}
+    {label}
+    onChange={setOptionValue}
+    {options}
+  />
```
cerebrl, line 62: Confirmation callbacks are almost always rendered as buttons, but this complicates how the step is rendered. One then has to figure out how to determine if the standard "Next" button should be rendered or not, breaking the principle of isolation.

To combat this, I added a select that provides the user the ability to select the confirmation value via a ... well, `select`, which removes the complexity of how to render all the buttons. This gives us some breathing room to make the UX decision down the line a bit.

src/lib/journey/callbacks/text-output/text-output.mock.ts
```diff
@@ -1,5 +1,67 @@
...
+            'Dangerous script tag should be rendered as plaintext: <br/> <script>alert("pwned")</script>',
+        },
+        {
+          name: 'messageType',
+          value: '0',
+        },
+      ],
+    },
+    {
+      type: CallbackType.SuspendedTextOutputCallback,
```
cerebrl, line 47: Testing both types of text-output-callbacks.

.svelte-kit/tsconfig.json
```diff
@@ -1,45 +1,31 @@
...
+    "../src/**/*.svelte",
+    "../src/**/*.js",
+    "../src/**/*.ts",
+    "../src/**/*.svelte",
+    "../tests/**/*.js",
+    "../tests/**/*.ts",
+    "../tests/**/*.svelte"
+  ],
+  "exclude": ["../node_modules/**", "./[!ambient.d.ts]**"]
+}
```
cerebrl, line 31: These changes come from SvelteKit, so it's managed by them.

cerebrl (approved):

PR #20, opened 2022-09-29
feat(logo): add logo support

cerebrl:
Jira: https://bugster.forgerock.org/jira/browse/SDKS-2050

## Summary

This is to add support for customer logos to be included at the top of the modal dialog. This feature will not be included with the inline form factor as the customer can fully control the surrounding UI. There are two ways to add a logo:

1. Simple logo or icon above the form's title, no different than the generic icon that's displayed now
2. A special modal header section created at the top of the modal for the logo to be displayed in more complex treatments

Currently, if there's the logo configuration present, we suppress the generic icon usually shown from the journey component. This removes too many images being shows at once. We could change this, but it would take some really good design treatment to preserve both the journey icon and display the customer logo in a cohesive way.


src/lib/components/compositions/dialog/dialog.svelte
```diff
@@ -41,23 +44,54 @@
 </script>
 
 <!-- Inspired by https://github.com/mvolkmann/svelte-dialog -->
-<!-- TODO: Animate modal from bottom and fade in -->
 <dialog
   id={dialogId}
   bind:this={dialogEl}
-  class={`tw_dialog-box dark:tw_dialog-box_dark md:tw_dialog-box_medium`}
+  class={`tw_dialog-box dark:tw_dialog-box_dark md:tw_dialog-box_medium ${
+    forceOpen ? '' : 'tw_dialog-box_animate'
```
cerebrl, line 51: This just eliminates the animation if the dialog is forced to be open for testing purposes. In the future, we can use this turn off animation if a user has reduced motion system preference enabled.

src/lib/components/compositions/dialog/dialog.svelte
```diff
@@ -41,23 +44,54 @@
...
-  <div class="tw_w-full tw_-mt-4 tw_relative tw_text-right">
-    <button
-      class="tw_dialog-x tw_focusable-element dark:tw_focusable-element_dark"
-      on:click={() => closeDialog({ reason: 'user' })}
-      aria-controls={dialogId}
-    >
-      <XIcon
-        classes="tw_inline-block tw_fill-current tw_text-secondary-dark dark:tw_text-secondary-light"
-        ><T key="closeModal" /></XIcon
+  {#if withHeader}
```
cerebrl, line 55: Because the order of HTML is different, and styles are quite different, it was just easier to split the HTML by a simple condition.

src/lib/components/compositions/dialog/dialog.svelte
```diff
@@ -41,23 +44,54 @@
...
+        class="tw_dialog-x md:tw_dialog-x_medium tw_focusable-element dark:tw_focusable-element_dark"
+        on:click={() => closeDialog({ reason: 'user' })}
+        aria-controls={dialogId}
       >
-    </button>
+        <XIcon
+          classes="tw_inline-block tw_fill-current tw_text-secondary-dark dark:tw_text-secondary-light"
+          ><T key="closeModal" /></XIcon
+        >
+      </button>
```
cerebrl, line 72: After looking at this, we could probably separate this out into it's own x-button component. Not sure I want to take this on at this moment though.

src/lib/components/compositions/input-stacked/stacked-label.svelte
```diff
@@ -27,7 +27,7 @@
   }
 </script>
 
-<div class="tw_input-spacing">
+<div class="tw_input-spacing tw_flex tw_flex-wrap">
```
cerebrl, line 30: Fixing a bug when the password component is used with stacked labels, rather than floating.

src/lib/components/primitives/input/input.svelte
```diff
@@ -10,7 +10,7 @@
   export let labelClasses = '';
   export let labelOrder: 'first' | 'last' = 'first';
   export let onChange: (event: Event) => void;
-  export let placeholder: string | null = null;
+  export let placeholder: string | null = ' ';
```
cerebrl, line 13: Literal space is needed as a default value in order for the floating label to work when we are not provided a placeholder. This was uncovered with the use of stacked labels.

src/lib/components/primitives/input/input.svelte
```diff
@@ -42,7 +42,7 @@
     class={`${inputClasses} tw_input-base dark:tw_input-base_dark tw_focusable-element dark:tw_focusable-element_dark tw_flex-1 tw_w-full`}
     id={key}
     on:change={onChange}
-    placeholder={placeholder || label}
+    {placeholder}
```
cerebrl, line 45: Removing the `label` fallback, and use the literal space character above, so that we don't end up with a duplication of label and placeholder.

src/lib/journey/journey.svelte
```diff
@@ -20,14 +22,21 @@
...
 {#if !$journeyStore?.completed}
-  <svelte:component
-    this={mapStepToStage($journeyStore.step)}
-    failureMessage={$journeyStore.error?.message}
-    bind:formEl
-    loading={$journeyStore.loading}
-    {submitForm}
-    step={$journeyStore.step}
-  />
+  {#if !$journeyStore.step}
```
cerebrl, line 25: This was moved here from the stage components. It reduces some of the duplication and simplifies the stage components a bit.

src/lib/journey/stages/generic.svelte
```diff
@@ -41,36 +41,38 @@
...
+  <!-- TODO: Needs localization strategy -->
+  <Sanitize html={true} string={step?.getHeader() || ''} />
+</h1>
+<p
+  class="tw_text-center tw_-mt-5 tw_mb-2 tw_py-4 tw_text-secondary-dark dark:tw_text-secondary-light"
+>
+  <!-- TODO: Needs localization strategy -->
+  <Sanitize html={true} string={step?.getDescription() || ''} />
+</p>
+
```
cerebrl, line 59: Trying to "flatten" this component out a bit. Moving some of the conditionals (checking for `step`) into the journey component allowed for this.

src/lib/style.store.ts
```diff
@@ -0,0 +1,26 @@
...
+    icon: boolean;
+  };
+}
+
+export let style: Readable<Style>;
+
+// TODO: Implement Zod for better usability
+export function initialize(customStyle: Style) {
+  style = readable(customStyle);
+}
```
cerebrl, line 26: This file will be the store for all style controls of the Widget. This allows us to not have to pass a ton of props around the app.

src/lib/components/compositions/dialog/dialog.svelte
```diff
@@ -1,10 +1,13 @@
 <script lang="ts">
   import T from '$components/_utilities/locale-strings.svelte';
   import XIcon from '../../icons/x-icon.svelte';
+  import { style } from '$lib/style.store';
 
   export let closeCallback: (args: { reason: 'auto' | 'external' | 'user' }) => void;
   export let dialogEl: HTMLDialogElement | null = null;
   export let dialogId: string;
+  export let forceOpen: boolean = false;
+  export let withHeader: boolean = false;
```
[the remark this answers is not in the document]
daveadams56: Is forceOpen primarily for testing purposes, or is it a setting that the developer may set in their implementation? If so can you explain a little more about when they would use this?

src/lib/components/compositions/dialog/dialog.svelte
daveadams56: Makes sense

src/lib/components/compositions/dialog/dialog.svelte
daveadams56: Just for my own knowledge, what do you mean by `x-button` and why / when would we consider refactoring this to an `x-button`?

daveadams56 (approved):
This looks good to me - I've taken a look at Chromatic and approved there, and left a couple of comments here. Nothing blocking, just a couple of things I'd like to understand a little better, if you have time to respond. One last thought - do we think any of the conflicts will be an issue, and do you plan to resolve these once we approve, or do you want to resolve then I'll re-review?

daveadams56 (approved):
Took a quick look around following your latest push - all looks good to me. One thing I've noticed just now and looking at the other PR for SDKS-2006, is that the demo app seems broken due to a CORS error. Having said that, I suspect the issue is that the URL that the PR lists towards the bottom is not added to the CORS configuration on the server. If we can quickly verify that this or a similar cause is behind the issue, then the PR is approved!

PR #21, opened 2022-10-04
fix(readme): address feedback from SDKS-2070

cerebrl:
Addresses findings provided in https://bugster.forgerock.org/jira/browse/SDKS-2070


daveadams56 (approved):
Excellent, I had another read through and this looks correct now. Approved!

PR #22, opened 2022-10-05
feat(Sdks 2001): add hidden value callback and radio display type for choice callbacks

daveadams56:
Jira: https://bugster.forgerock.org/jira/browse/SDKS-2001

## Summary

This PR adds the support for `HiddenValueCallback` and extends the support for the `ChoiceCallback` with the option to display the choices as radio buttons. The latter needs further work before this can be merged, given the ongoing discussion about supporting callback metadata in the `stage` parameter since this is how AM specifies that the callback should be displayed as **radio** or **select**.


src/lib/journey/callbacks/choice/choice.svelte
```diff
@@ -17,6 +19,9 @@
   const choiceOptions = callback
     .getChoices()
     ?.map((label, idx) => ({ text: label, value: `${idx}` }));
+  const choiceOptionsForRadio = callback
+    .getChoices()
+    ?.map((label, idx) => ({ text: label, value: idx }));
```
daveadams56, line 24: I had to duplicate `choiceOptions` here since the Radio composition expects the choices to be formatted as `{ text: string, value: number } `as opposed to `{ text: string, value: string }` which is what Select expects. I would welcome suggestions here for how we can improve this.

daveadams56, line 24: I had to duplicate `choiceOptions` here since the Radio composition expects the choices to be formatted as `{ text: string, value: number } `as opposed to `{ text: string, value: string }` which is what Select expects. I would welcome suggestions here for how we can improve this.

src/lib/journey/callbacks/choice/choice.svelte
```diff
@@ -37,13 +42,26 @@
...
+{:else}
+  <Radio
+    {firstInvalidInput}
+    defaultOption={Number(defaultChoice)}
+    isRequired={false}
+    key={inputName}
+    message={prompt}
+    onChange={setValue}
+    name={inputName}
+    options={choiceOptionsForRadio}
```
daveadams56, line 65: See above comment about the `choiceOptions`

src/lib/journey/callbacks/hidden-value/hidden-value.mock.ts
```diff
@@ -1,3 +1,31 @@
...
+    {
+      type: CallbackType.HiddenValueCallback,
+      output: [
+        {
+          name: 'value',
+          value: 'false',
+        },
+        {
+          name: 'id',
+          value: 'webAuthnOutcome',
```
daveadams56, line 16: Taking a quick look at the AM codebase, I can see that the following nodes could return a HiddenValueCallback:

- PushWait
- PushRegistration
- WebAuthn
- WebAuthn Registration

... and of course any scripted nodes that could return one.

I chose WebAuthn here, though it will be difficult to test this callback properly and be confident we have implemented everything, until we implement support for these other node callbacks.

src/lib/journey/callbacks/hidden-value/hidden-value.svelte
```diff
@@ -0,0 +1,11 @@
...
+  import type { HiddenValueCallback } from '@forgerock/javascript-sdk';
+
+  export let callback: HiddenValueCallback;
+  export let idx: number;
+
+  const inputName = callback?.payload?.input?.[0].name || `hidden-value-${idx}`;
+  const inputValue = callback?.getInputValue();
+</script>
+
+<input type="hidden" name={inputName} id={inputName} value={inputValue} />
```
daveadams56, line 11: As you can see we're not rendering much here - but I think it's important that we do render something, so that scripts could be supported in future.

src/lib/journey/callbacks/choice/choice.svelte
```diff
@@ -37,13 +42,26 @@
...
+    onChange={setValue}
+    options={choiceOptions}
+  />
+{:else}
+  <Radio
+    {firstInvalidInput}
+    defaultOption={Number(defaultChoice)}
+    isRequired={false}
+    key={inputName}
+    message={prompt}
```
daveadams56, line 62: I found that the formatting of Radio is odd, with the choices first, followed by the prompt. It looks like this is intentional, but I'm not sure why. @cerebrl are you able to explain a little about this? Personally, I think it should be the other way round, but want to understand why it is how it is now. See screenshot below for example:

Screenshot 2022-10-05 at 18 56 19

daveadams56:
This PR is not intended to be merged just yet - there is still some work to do, I feel. Please review when you get time, and let's discuss. Thanks.

src/lib/journey/callbacks/choice/choice.svelte
cerebrl: Ah, this is a good callout. There are two things here:

1. You're using `message` prop for displaying the `prompt`, this is why it's displayed below
2. Both the standard and the animated radio don't have "group" label, which is needed here to have something describing the group of radios available for selection.

For (1) above, this is for adding informational or error messages to an input, like input requirements or validation errors. For (2), we need to actually add this to the underlying component. Since the components already have the concept of a label for the input (the "Yes" and "No" in the above screenshot), we need to add a "group label" to the component at the composition level.

So, for `animated.svelte` and `standard.svelte`, let's add a "grouping" header, so to speak, as an exported prop that you can use to pass in the `prompt` property. I'm most familiar with using `fieldset` for doing this kind of grouping, but we may have other options. I'm mostly concerned with the accessibility of our choice. You can read more about it here: https://www.w3.org/WAI/tutorials/forms/grouping/.

You can use basic tailwind utility classes to control the styling: https://tailwindcss.com/docs/utility-first. You don't have to worry about adding a custom class to the theme for this right now.

src/lib/journey/callbacks/choice/choice.svelte
cerebrl: I see this as a type bug. Let's fix it, so we don't have to duplicate the data structure. Honestly, the actual DOM treats the value as a string, regardless of what you pass in, so typing it as a number is a bit misleading. This change will have to be done for the primitive and the two compositional components (animated and standard).

src/lib/journey/callbacks/choice/choice.svelte
cerebrl: Let's fix the type bug, so we don't have to do this. Let me know if that makes sense.

src/lib/journey/callbacks/hidden-value/hidden-value.mock.ts
cerebrl: It's worth noting that the SDK doesn't actually execute the script type from the WebAuthn nodes. [It parses the script to extract the values](https://github.com/ForgeRock/forgerock-javascript-sdk/blob/master/packages/javascript-sdk/src/fr-webauthn/script-parser.ts) and then internally handles it like the JSON version. 

As for the Push nodes, this is for the QR code display. I'd rather "natively" support QR codes, rather than have the Widget actually run the external script. So, we will either parse the script like the WebAuthn or add a switch to send JSON rather than a script from the node.

My long term vision for the Widget is for us to educate the customer on adding native support for the features they want by extending the Widget, rather than writing the scripts in AM and having the Widget execute them. From an architectural standpoint, supporting the features in the Widget is far superior from a quality control and security standpoint. 

I would highly recommend we treat the execution of scripts from AM within client apps to be a legacy method, and provide this superior method and habituate customers to it, discouraging the "legacy" method. With that said, I would actually be okay with this being a "noop" component as we just need a component to exist, so Svelte doesn't get tripped up.

From a WebAuthn callback/feature perspective, the SDK just directly sets the value on the callback internally, and NOT through any hidden DOM input.

src/lib/journey/callbacks/hidden-value/hidden-value.svelte
cerebrl: Understood. Honestly, I'm struggling with supporting this "legacy" style of handling customization. I touched on this above, but I'll reiterate it here as I think this is really important for us to seriously consider.

If you think about the _core_ reason for this pattern – sending scripts from AM for a client to execute in the DOM and injecting the side-effect of this execution into a hidden input – it's all about customizing the client's behavior when you _don't have access to the source code_. 

This was always true for XUI and is still somewhat true for Platform Login. Customers don't have a holistically designed process for customizing or adding features to a login flow that is not natively supported by the client application, and they can't easily alter said application. 

The Login Widget is directly designed to address this. We are explicitly designing a framework that allows them to add, extend, customize the login flow directly in the client application. Through good design, education and evangelism, I feel we can effectively "flip" the mental model of how our customers control their UX on its head and provide a far superior developer experience while doing so.

Anyway, I'm happy to continue this conversation with you and the rest of the team. I think it's vital for us to deeply understand this concept and evaluate this feature without the historical bias of its legacy. If, after deep contemplation, we still have a strong reason to allow for this "legacy" (as I call it) feature, then I'm happy to support it :)

daveadams56: I understand your point of view - and I agree, when it comes to allowing scripting for the sake of modifying the UI. However my concern is related to the partner integrations - the nodes of some of these return a client side script to render something and may expect a hidden value callback to render something on the page to capture the result. I think for this change, perhaps we can decide to render nothing, however I think it merits further discussion.

src/lib/journey/callbacks/choice/choice.svelte
daveadams56: Understood, I think I know what I need to do for this. I'll have a go at this and hopefully push my changes later today. Thanks for the guidance.

src/lib/journey/callbacks/hidden-value/hidden-value.mock.ts
daveadams56: I think I understand, and aside from the concern in the other comment about partner nodes, I'm happy that we don't need to render anything. 

To clarify, when you say the following:

> With that said, I would actually be okay with this being a "noop" component as we just need a component to exist, so Svelte doesn't get tripped up.

... Are you saying that you think we should render something anyway, or just that we have a Svelte component but don't render anything?

src/lib/journey/callbacks/choice/choice.svelte
daveadams56: I've implemented a grouping header as per w3 above - @cerebrl please take another look. Thanks.

src/lib/journey/callbacks/choice/choice.svelte
daveadams56: This was easy enough to fix I think - @cerebrl please take another look. Thanks.

src/lib/journey/callbacks/hidden-value/hidden-value.mock.ts
cerebrl: I'm actually not certain what the minimum requirement is for a Svelte component. Is a Svelte component valid with no HTML? I honestly don't know. If it needs some kind of HTML, we could just render an empty span tag?

From a type perspective, I would think it at least needs the same required, exported props, as the rest of the callback components to ensure the APIs line up. In other words, these:

1. `callback`
2. `firstInvalidInput`
3. `idx`

src/lib/journey/callbacks/hidden-value/hidden-value.svelte
cerebrl: Absolutely agree with discussing this further as I'd love to know more about these partner integrations you mention.

src/lib/journey/callbacks/hidden-value/hidden-value.mock.ts
daveadams56: Hmm good point - as far as I can tell, a Svelte component can be empty and still be valid. My assumption is based on an [article](https://dev.to/jesseskinner/the-simplest-svelte-component-is-an-empty-file-2lhk) and something I saw in the [docs](https://svelte.dev/docs):

> All three sections — script, styles and markup — are optional.

I tested this theory by refactoring the hidden value component to the following, and didn't get yelled at by my IDE, storybook or the test app!

src/lib/journey/callbacks/choice/choice.svelte
```diff
@@ -1,11 +1,13 @@
 <script lang="ts">
   import type { ChoiceCallback } from '@forgerock/javascript-sdk';
 
+  import Radio from '$components/compositions/radio/animated.svelte';
   import Select from '$components/compositions/select-floating/floating-label.svelte';
 
   export let callback: ChoiceCallback;
   export let firstInvalidInput: boolean;
   export let idx: number;
+  export let displayType: 'radio' | 'select' = 'select';
```
daveadams56, line 10: Given that we have decided (for now at least) not to react to the content in the stage parameter that drives this in a journey / tree, I've hard coded this to `select`. I'm fine with this, as this may be a capability we wish to introduce later, however for now I'm struggling to understand how we should leave this with regards to testing and review in Storybook / Chromatic. We can't effectively specify `radio` or `select` so all we will see at review is the effect of `select`. Hope this makes sense - @cerebrl any ideas?

src/lib/journey/callbacks/choice/choice.svelte
cerebrl: This looks great!

src/lib/journey/callbacks/hidden-value/hidden-value.mock.ts
cerebrl: Perfect. Let's keep this dead simple, and make decisions about any additional handling when they come up in testing.

src/lib/journey/callbacks/choice/choice.svelte
cerebrl: This changes look good. I have some comments on naming, but I'll resolve this conversation and add my specific comments inline there.

src/lib/journey/callbacks/choice/choice.svelte
cerebrl: Yeah, let's go ahead and fully implement and test this `displayType` as a feature. Let's just keep it as a manual option that can be set in a custom "named" stage component. We just won't wire it up to the Stage property coming from AM until we settle on our response to the new news. 

So, for Storybook, we will have the Choice item with "Base" as the select type (since it is the default value), and then we will have a variant that uses the radio type. You can have a super simple Storybook interaction test like we have on the ConfirmationCallback: https://github.com/cerebrl/forgerock-web-login-framework/blob/main/src/lib/journey/callbacks/confirmation/confirmation.stories.js#L43.

As for e2e tests, I'd create a new journey by copying `LoginWithConfirmation` and adding a Choice node to it and rename it. Then modify this test by pointing it to the new journey and accommodating the new node/callback: https://github.com/cerebrl/forgerock-web-login-framework/blob/main/tests/widget/modal/widget-modal.login-with-confirm.test.js (rename it too). I say that because I want to avoid too many shallow e2e tests hitting our tenant in parallel. Just adding a node should be a more stable way to include it in our e2e without exponentially increasing the stress on the server.

src/lib/components/compositions/radio/animated.svelte
```diff
@@ -5,15 +5,16 @@
   import type { Maybe } from '$lib/interfaces';
   import Label from '$components/primitives/label/label.svelte';
 
-  export let defaultOption: number;
+  export let defaultOption: string | null = null;
   export let message = '';
   export let firstInvalidInput: boolean;
+  export let groupMessage: string;
```
cerebrl, line 11: To keep the naming consistent with the rest of the components, I think calling this a `groupLabel` would be better than `groupMessage`. I say that since labels (as a convention in this project) generally go above the input and messages below.

src/lib/components/compositions/radio/animated.svelte
```diff
@@ -27,35 +28,38 @@
...
-        classes="tw_input-spacing tw_grid tw_grid-cols-[2.5em_1fr] tw_relative"
-      >
-        <span class="tw_animated-radio dark:tw_animated-radio_dark" />
-        {option.text}
-      </Label>
-    </div>
-  {/each}
-  <!--
+<fieldset>
+  <legend>{groupMessage}</legend>
```
cerebrl, line 32: Cool, I'll pull this in locally and run it to see how it looks.

src/lib/journey/_utilities/map-callback.utilities.ts
```diff
@@ -10,6 +10,7 @@ import Name from '$journey/callbacks/username/name.svelte';
 import Password from '$journey/callbacks/password/password.svelte';
 import StringAttributeInput from '$journey/callbacks/string-attribute/string-attribute-input.svelte';
 import TermsConditions from '$journey/callbacks/terms-and-conditions/terms-conditions.svelte';
+import HiddenValue from '$journey/callbacks/hidden-value/hidden-value.svelte';
```
cerebrl, line 13: Would you mind moving this to alphabetize it according to the filename? So, between `confirmation.svelte` and `kba-create.svelte`. Most things are alphabetized by variable, function or property name, but imports can get weird with default and named imports, along with single and multiple imports, so going be filename just makes it easier.

cerebrl (changes requested):
Left a few stylistic/semantic comments. The only substantial thing left is the accommodation of the new node/callback into an e2e test, and we're good.

cerebrl:
Oh, I've just noticed that your PR is coming from your personal fork. Currently, our CI doesn't work with forks, so I'd like to ask that you push your branch to this repo, and create a new PR from your branch to `main` so the CI will run. You should be able to push as you are a contributor to this repo, but let me know if you have any issues. Once the new PR is open, just reference this PR from the new one and you can close this out.

cerebrl:
It also looks like you have some merge conflicts, so if you could rebase, that would be great.

src/lib/components/compositions/radio/animated.svelte
cerebrl: Pulled code locally, added radio as a variant of Choice in the `choice.stories.js` file, and ran Storybook. Looks like the group label needs just a touch of styling and support of dark mode.

Screenshot:
https://share.getcloudapp.com/eDu77jBG

Directly adding some tailwind classes will do the trick. These will be sufficient:

```
<legend class="tw_input-label dark:tw_input-label_dark tw_font-bold tw_mb-4">
```

I also observed that the render for the Composition story for radio is rendering `undefined`, so we should probably handle the falsey value by defaulting to empty string if the exported prop is not passed.

Screenshot:
https://share.getcloudapp.com/2Nu77wzO

This will do it:

```
export let groupLabel = '';
```

That way, when nothing is passed, it doesn't render `undefined`.

daveadams56: Ah good spot. Thanks.

src/lib/journey/_utilities/map-callback.utilities.ts
daveadams56: Absolutely. Makes sense.

src/lib/components/compositions/radio/animated.svelte
daveadams56: Makes sense to me. Will do.

src/lib/journey/callbacks/choice/choice.svelte
daveadams56: I've been looking at Storybook and am part of the way there I think.. We don't have an interaction test for "Base" choice i.e. select, so do we want an interaction for each of these?

cerebrl: For Storybook Interactions, we need to test anything that contains some sort of view/rendering related logic, or if the component interfaces with another module, like the SDK. If the component just plainly renders something, then no need to do an interaction test on that, since there's no "interaction" per se (Testing whether some just renders is testing whether Svelte works, and we can trust that Svelte has its own tests for that).

Since this component does interact with the SDK, the minimal test is just answering the question of whether changing the value in the DOM input also changes the value in the expected callback `input` field. That's all, especially when the component is so simple. No elaborate test is needed.

daveadams56: Understood. I've added one for "base" and another for "radio", given we have two variants for choice both interacting with the SDK. Thanks

src/lib/journey/callbacks/choice/choice.stories.js
```diff
@@ -21,3 +26,54 @@ export const Base = {
...
+
+  await new Promise((resolve) => setTimeout(resolve, 1000));
+
+  await userEvent.selectOptions(select, '1');
+  await expect(cb.getInputValue()).toBe(1);
+};
+
+export const RadioInteraction = Template.bind({});
+
+RadioInteraction.args = { ...Radio.args };
```
daveadams56, line 64: In response to a previous conversation we had, I decided to add two interaction tests here - one for choice / select, and another for choice / radio. I think this is the only place we have multiple interactions for the same component - I'd welcome your thoughts on this approach generally, naming (i.e. Interaction vs RadioInteraction), and if there is a more concise way to achieve this.

package.json
```diff
@@ -9,6 +9,7 @@
     "chromatic": "chromatic --project-token=e10acf0c74f9 --force-rebuild",
     "ci": "vitest --dir src",
     "ci:e2e": "playwright test",
+    "debug:ci:e2e": "PWDEBUG=1 playwright test",
```
cerebrl, line 12: Can you reverse the wording here? I'd like the order to go from lower to higher specificity. So, this instead: `ci:e2e:debug`.

src/lib/components/compositions/radio/animated.svelte
```diff
@@ -5,15 +5,16 @@
   import type { Maybe } from '$lib/interfaces';
   import Label from '$components/primitives/label/label.svelte';
 
-  export let defaultOption: number;
+  export let defaultOption: string | null = null;
```
cerebrl, line 8: Can we use the `Maybe` here for consistency? So, `Maybe<string>`, rather than `string | null`?

src/lib/journey/callbacks/choice/choice.stories.js
cerebrl: I think this is fine.

cerebrl (changes requested):
I think this looks good. I just had two small comments. Please close this PR and open a new one with a branch directly from this repo, rather than a branch from your personal fork. This will allow the CI to work appropriately. Then, just close this and reference from your other PR. Thanks.

daveadams56:
Closing this PR as we now have a new one for these changes, within the login widget repo - [PR 26](https://github.com/cerebrl/forgerock-web-login-framework/pull/26).

PR #24, opened 2022-10-18
feat(callbacks): display link to terms and conditions

daveadams56:
Jira: https://bugster.forgerock.org/jira/browse/SDKS-2004

## Summary
This PR extends the support for `TermsAndConditionsCallback` with the ability a configurable URL to the terms and conditions which are hosted separately.


src/lib/journey/callbacks/terms-and-conditions/terms-conditions.svelte
```diff
@@ -40,5 +47,8 @@
 </script>
 
 <Checkbox {firstInvalidInput} key={inputName} onChange={setValue} value={false}>
-  <T key="termsAndConditions" />
+  <span>
```
cerebrl, line 50: Is the `span` necessary?

src/lib/components/primitives/message/input-message.svelte
```diff
@@ -1,11 +1,14 @@
 <script lang="ts">
+  import sanitize from 'xss';
   import type { Maybe } from '$lib/interfaces';
 
   export let message: string;
```
cerebrl, line 5: I'd like to change the prop's name here from `message` to `dirtyMessage` (I know that sounds silly, but I like being obvious when security is involved). If you use the VS Code "Rename Symbol" function (right click), it should change all references automatically for you.

cerebrl (commented):
Looking good so far. Just a few comments.

src/lib/journey/callbacks/terms-and-conditions/terms-conditions.svelte
daveadams56: I had hoped not, but it seems to affect the formatting of the link if I remove it:

Screenshot 2022-10-19 at 12 59 48

Maybe there's something we can do with CSS for this if you think it's preferable to removing the span?

src/lib/components/primitives/message/input-message.svelte
daveadams56: Makes sense.

src/lib/journey/callbacks/terms-and-conditions/terms-conditions.svelte
cerebrl: Oh, yeah, hmm ... let's add an optional `classes` prop to the `Message` component, so we can pass in a few extra classes for optional styling purposes. Here are the classes you'd want to pass in:

```
tw_col-start-2 tw_row-start-2
```

These classes just instruct the `Message` component to render in a different block of the grid used for the `Checkbox` component.

src/lib/journey/callbacks/terms-and-conditions/terms-conditions.story.svelte
```diff
@@ -1,10 +1,14 @@
...
   import type { TermsAndConditionsCallback } from '@forgerock/javascript-sdk';
+  import { initialize } from '$lib/links.store';
 
   import Centered from '$components/primitives/box/centered.svelte';
   import Terms from './terms-conditions.svelte';
 
   export let callback: TermsAndConditionsCallback;
+
+  // Initialize links
+  initialize({ termsAndConditions: 'https://www.forgerock.com/terms' });
```
daveadams56, line 11: I chose the ForgeRock terms page as the default / test URL for Storybook, e2e and test apps. Open to suggestions on what the "default" within the project should be if there are any.

src/lib/links.store.ts
```diff
@@ -0,0 +1,11 @@
+import { readable, type Readable } from 'svelte/store';
+
+export interface Links {
+  termsAndConditions?: string;
+}
+
+export let links: Readable<Links>;
```
daveadams56, line 7: I implemented the store more like the style.store instead of locale.store, since it feels closer to this in nature - i.e. reading some extra props we define in the widget config, instead of the more advanced behaviour of locale.store.

This means that we don't export something similar to `partialStringsSchema` as the locale store does. So this differs slightly from the suggestion @cerebrl made in our chat on slack, but for me this is a simple implementation that works and is consistent with other stores. Having said that I welcome feedback on my approach during the review.

src/lib/components/primitives/message/input-message.svelte
```diff
@@ -1,11 +1,15 @@
 <script lang="ts">
+  import sanitize from 'xss';
   import type { Maybe } from '$lib/interfaces';
 
-  export let message: string;
+  export let classes = '';
```
daveadams56, line 5: This was a really simple solution to eliminate the `<span>` that was picked up on earlier - thanks @cerebrl 👍

daveadams56:
I still need to implement e2e tests for this - I haven't figured out yet how to verify the contents of a new tab in a test using playwright.

src/lib/links.store.ts
cerebrl: Unfortunately, the `style.store` is incomplete. The `locale.store` is the best example as it's complete in its implementation. The difference is the use of Zod. Zod provides both build-time type safety, _plus_ runtime type feedback, which is critical to help developers catch silly misspellings and such. 

This is important for config and instantiation (public APIs) as it will eliminate those hard to catch bugs from improper object structure or typos in object keys. The convention is to use Zod for public APIs as devs may not always use TypeScript well or at all. We don't heavily use it internally, but this may change over time.

cerebrl (commented):
First scan through this looks pretty good. Left a few comments. I will try to find time to pull this in locally and run through it.

rodrigoareis (commented):
Changes looks good to me

src/lib/links.store.ts
daveadams56: Ok, this makes sense to me now. I've updated the links store to use Zod. Thanks.

src/lib/journey/callbacks/terms-and-conditions/terms-conditions.svelte
```diff
@@ -21,7 +25,9 @@
    ************************************************************************* */
   const Checkbox = checkAndRadioType === 'standard' ? Standard : Animated;
   const inputName = callback?.payload?.input?.[0].name || `terms-${idx}`;
-  const terms = callback.getTerms();
+  const termsLinkHtml = `<a href=${$links?.termsAndConditions} target="blank">${interpolate(
```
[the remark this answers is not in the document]
daveadams56: Good spot!

tests/widget/modal/widget-modal.register.test.js
```diff
@@ -17,6 +17,14 @@ test('modal widget', async ({ page }) => {
   await page.fill('text=Password', 'j56eKtae*1');
   await page.selectOption('select', '0');
   await page.fill('text=Security Answer', 'Red');
+
+  const termsLink = page.locator('text=View full Terms & Conditions');
+  const termsUrl = await termsLink.getAttribute('href');
+  const termsTarget = await termsLink.getAttribute('target');
+
+  expect(termsUrl).toBe('https://www.forgerock.com/terms');
+  expect(termsTarget).toBe('_blank');
```
daveadams56, line 26: On the review call for this PR, we discussed approaches for verifying that the terms link **a)** is correct / as configured and **b)** opens in a new tab. I originally set out to use Playwright to actually click the link and verify that this opens in a new tab. However after discussing this with the team, the consensus seems to be that the value in the e2e test is verifying that the URL is displayed as configured and _should,_ open in a new tab, rather than that the link actually opens in a new tab. I personally think this approach is sufficient, but welcome feedback.

src/lib/links.store.ts
```diff
@@ -0,0 +1,14 @@
+import { readable, type Readable } from 'svelte/store';
+import { z } from 'zod';
```
daveadams56, line 2: As requested @cerebrl here is the links store re-written using Zod. This actually seemed quite simple to do. Let me know what you think.

src/lib/journey/callbacks/terms-and-conditions/terms-conditions.svelte
```diff
@@ -21,7 +25,9 @@
    ************************************************************************* */
   const Checkbox = checkAndRadioType === 'standard' ? Standard : Animated;
   const inputName = callback?.payload?.input?.[0].name || `terms-${idx}`;
-  const terms = callback.getTerms();
+  const termsLinkHtml = `<a href=${$links?.termsAndConditions} target="_blank">${interpolate(
```
cerebrl, line 28: What should we do in cases where the `links.termsAndConditions` is not provided? As it is now, I feel like this could lead to hidden bugs since we are not conditionally rendering the link on the presence of the URL. The result would be a visually correct, but non-functional link.

Just "thinking out loud" here, but should we render an error if the URL is not provided? That way when the developer is testing, it's obvious the URL is missing. Or, should we just not rendering the message at all? Hmm ...

cerebrl (commented):
I think this is good. I have two questions and the involve UX and DX questions:

1. What do we do when the dev doesn't pass in the URL for link creation?
2. Do we render an icon to denote opening in a new tab?

Let's discuss this on Friday.

cerebrl:
Oh, @daveadams56, make sure to address the merge conflicts. Thanks.

src/lib/journey/callbacks/terms-and-conditions/terms-conditions.svelte
daveadams56: Good point. I think we should render an error if no URL is provided, rather than silently displaying nothing. It feels like there are risks involved if we decide to display nothing and this slips through to a customer's deployed app, since users will not have a chance to read what they are accepting, potentially leading to legal issues. Let's discuss further on our call today as you say.

src/lib/journey/callbacks/terms-and-conditions/terms-conditions.svelte
```diff
@@ -41,4 +47,5 @@
 
 <Checkbox {firstInvalidInput} key={inputName} onChange={setValue} value={false}>
   <T key="termsAndConditions" />
+  <Message classes="tw_col-start-2 tw_row-start-2" dirtyMessage={termsLinkHtml} />
```
cerebrl, line 50: Just a reference to a good resource about this a11y concern that led to not placing the link in label: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#accessibility_concerns.

ryanbas21 (approved):
this looks good to me

PR #25, opened 2022-10-18
feat(self-submitting): add self-submitting form and pollingwait support

cerebrl:
## Jira

Ticket: https://bugster.forgerock.org/jira/browse/SDKS-1999

## Summary

There are certain callbacks from AM that represent "self-submittable" callbacks. In other words, when the user makes a selection or a period of time has elapsed, the callback can immediately be submitted without the need for explicit submission by user. This is in contrast to "ordinary" callbacks that simply request user information and requires the user to explicitly click submit.

## Callbacks added

1. `PollingWaitCallback`
2. `ConfirmationCallback` (partial support existed prior)

## Accessibility

This PR also introduces better accessibility with improved focus management for multi-step journeys.

## Details

The central functionality for the self-submitting callbacks is a reliance on callback and step metadata. This metadata is built up separate to the AM response that contains the step and callback information. This metadata aligns with the original callbacks in order for iteration alignment to be possible.

Each callback component shares its callback metadata with the parent step callback, but the step metadata is copied and given to the callback component. This prevent accidental mutation. Callback components are not allowed to modify data/state outside of itself.


CONTRIBUTING.md
```diff
@@ -0,0 +1,162 @@
+# Contributing to the ForgeRock Web Login Framework
```
cerebrl, line 1: Just adding a CONTRIBUTING.md to this project and starting with some questions that I've been asked and the answers I gave. This will at least get us started with some content regarding the development side of the framework.

src/lib/components/compositions/checkbox/animated.story.svelte
```diff
@@ -38,11 +38,11 @@
 </script>
 
 {#if withForm}
-  <Form onSubmitWhenValid={submitForm}>
+  <Form ariaDescribedBy="animatedCheckboxStory" onSubmitWhenValid={submitForm}>
     <Checkbox
       bind:this={el}
       {checkValidity}
-      firstInvalidInput={false}
+      isFirstInvalidInput={false}
```
cerebrl, line 45: You'll see a lot of this. I'm trying to normalize the boolean metadata variables to start with `is` for consistency.

src/lib/components/compositions/checkbox/animated.story.svelte
```diff
@@ -38,11 +38,11 @@
 </script>
 
 {#if withForm}
-  <Form onSubmitWhenValid={submitForm}>
+  <Form ariaDescribedBy="animatedCheckboxStory" onSubmitWhenValid={submitForm}>
```
cerebrl, line 41: This new `ariaDescribedBy` prop is new, and I've added it, so we can link the headers, errors or messages to the form itself for better accessibility.

src/lib/journey/_utilities/metadata.utilities.ts
```diff
@@ -0,0 +1,32 @@
...
+  return step?.callbacks.map((callback, idx) => {
+    return {
+      isFirstInvalidInput: checkValidation(callback),
+      isReadyForSubmission: isCbReadyByDefault(callback),
+      isSelfSubmittingCb: isSelfSubmittingCb(callback),
+      isUserInputRequired: requiresUserInput(callback),
+      idx,
+    };
+  });
+}
```
cerebrl, line 32: This is one of the main features of the "self-submitting" functionality that I added to support the `PollingWaitCallback` as well as the `ConfirmationCallback`. The idea is to build up these metadata objects outside of the actual step data and its callbacks, and use them to help inform portions of the code of the current step's nature.

I think this will also set us up for the new use of the Stage attribute by Platform UI. We can extract the data out of that stage properly, and inject it here to inform the components of intent.

There are currently two structures here: 1) the callback metadata, which is an array that matches the official callback array, and 2) the step metadata, which is a summary of the step itself and what it contains.

src/lib/journey/stages/generic.svelte
```diff
@@ -26,56 +32,69 @@
...
   }
 
+  afterUpdate(() => {
+    alertNeedsFocus = !!failureMessage;
+  });
+
   $: {
+    checkValidation = initCheckValidation();
+    callbackMetadataArray = buildCallbackMetadata(step, checkValidation);
+    stepMetadata = buildStepMetadata(callbackMetadataArray);
```
cerebrl, line 58: Here's where I'm calling for the build-up of the metadata, but I'm thinking about moving this into the journey's store.

src/lib/journey/_utilities/metadata.utilities.ts
daveadams56: This approach makes sense. I pulled in your code to run it, and found that we now see a reasonable amount of noise in the console when we run the app - and quite a lot more noise when we build it. These are warnings similar to the below, since we are declaring props that we are not using - these come from the new metadata and functions we are injecting.

```
16:11:34 [vite-plugin-svelte] /Users/david.adams/Checkouts/forgerock-web-login-framework-review/src/lib/journey/callbacks/username/validated-create-username.svelte:9:11 Validated_create_username has unused export property 'stepMetadata'. If it is for external reference only, please consider using `export const stepMetadata`
 7: export let callbackMetadata;
 8: export let selfSubmitFunction = null;
 9: export let stepMetadata;
```
Is there a way to reduce / suppress this noise, or are we happy to live with these for now? Not a huge problem, just one that would be good to reduce if it's easy enough since it can be unnerving seeing so many warnings!

src/lib/journey/callbacks/confirmation/confirmation.mock.ts
```diff
@@ -36,7 +36,7 @@ export default {
       ],
     },
     {
-      type: CallbackType.ConfirmationCallback,
+      type: 'ConfirmationCallback',
```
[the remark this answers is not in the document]
daveadams56: It's just a mock so doesn't really matter! Just wondered that's all.

daveadams56 (approved):
I ran the code locally to test it against a tree with polling wait on my tenant - this works nicely. Also took a look at Storybook locally, and again this change looks good. None of my comments are blocking, but are more things I want us to think about.

daveadams56:
It would be nice if the preview app was able to show a tree that includes this callback, but given we just use the Login / Register trees, the preview app isn't useful for much. As we add more callbacks and development progresses, I can see this being more of an issue. Is this easy enough to do, and if so should we make sure that each PR contains a preview app with an appropriate tree?

src/lib/journey/_utilities/metadata.utilities.ts
cerebrl: Yeah, I intent to solve these warnings, but I just ran out of time as this was supposed to be a 3 point story ... it ended up being about an 8. I could add addressing this in the CI story, maybe? But, I agree, I hate a noisy console log too :)

src/lib/journey/callbacks/choice/choice.svelte
```diff
@@ -76,7 +81,7 @@
   />
 {:else}
```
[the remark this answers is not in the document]
cerebrl: Yes, this is possible now. We just have to determine what constitutes a "self-submittable" context for the choice callback.

src/lib/journey/callbacks/confirmation/confirmation.mock.ts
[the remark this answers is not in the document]
cerebrl: Oh, weird. I don't think this was intentional at all.

tests/widget/modal/widget-modal.login-with-misc-callbacks.test.js
```diff
@@ -14,14 +14,17 @@ test('Modal widget with simple login and confirmation', async ({ page }) => {
...
   // Choice
-  expect(await page.locator('text="Are you sure?"').innerText()).toBe('Are you sure?');
-  await page.selectOption('select', '0');
+  const selectEl = page.getByLabel('Are you sure?');
+  await selectEl.selectOption('0');
   await page.locator('button', { hasText: 'Next' }).click();
 
+  // Polling Wait
+  // NOTE: Make sure timer is same or more than set in Polling Wait node
+  await page.waitForTimeout(3000);
```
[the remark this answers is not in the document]
cerebrl: Ah, good point. I can absolutely increase it to the default.

cerebrl:
For testing, you can now pass in the `journey` URL parameter and provide a custom journey for use by the main app itself. For example: https://forgerock-web-login-framework-git-sdks-1999-featpoll-756b48-cix.vercel.app/?journey=LoginWithMiscCallbacks. That will render the journey best for testing multi-step journeys and the self-submittable callbacks.

ryanbas21 (approved):
i've read through this a bit, i dont have much to add though. the changes all make sense and I agree with the comments that Dave has left.

PR #27, opened 2022-10-19
test(sdks-2051/storybook UI tests): Add more storybook tests

ryanbas21:
Added more storybook interaction tests to the repo. was able to get the mock functions working with justins help.


src/lib/components/primitives/select/select.stories.js
```diff
@@ -54,3 +56,28 @@ export const LabelLast = {
...
+
+  await userEvent.tab();
+  expect(label).toHaveFocus()
+
+  await userEvent.selectOptions(label, ['0'])
+  expect(canvas.getByText('Red')).toBeTruthy()
+  await userEvent.selectOptions(label, ['1'])
+  expect(canvas.getByText('Green')).toBeTruthy()
+  await userEvent.selectOptions(label, ['2'])
+  expect(canvas.getByText('Blue')).toBeTruthy()
```
cerebrl, line 82: I'm unsure about the value being tested here. I do feel this is testing native browser behavior, but I also think the assertions will never fail. Selecting an option doesn't remove the other options text from the DOM, I could be wrong, but regardless, I think this may distract us from a more important feature of the component to test.

This component has custom logic that we wrote that is much more likely to fail than a native browser element. For example:

1) Let's make sure a `defaultValue` passed into the component is indeed selected appropriately when rendered

2) Make sure the auto-keyboard focus works as expected. See here:
https://github.com/cerebrl/forgerock-web-login-framework/blob/main/src/lib/components/primitives/select/select.svelte#L28

    The above logic means if this input is invalid, it must apply keyboard focus to itself as it's the first of a list of invalid inputs. This would be good to test.

3) We need to ensure our event handler and `onChange` function passed in all are called as expected. See here: https://github.com/cerebrl/forgerock-web-login-framework/blob/main/src/lib/components/primitives/select/select.svelte#L34

4) We also have the small bit of logic that handles if the `null` option has the same text as the label, it doesn't display the label as it's redundant.

In my opinion, the above have a much higher ROI as we are more likely to break the custom logic we have than a browser breaking how a `select` element works.

src/lib/components/primitives/label/label.stories.js
```diff
@@ -15,3 +17,17 @@ export const Base = {
...
+  Component: Label,
+  props: args
+});
+export const Interaction = Template.bind({});
+Interaction.args = { ...Base.args };
+Interaction.play = async ({ canvasElement }) => {
+  const canvas = within(canvasElement);
+  const text = canvas.getByText('Input Label')
+  expect(text).toBeInTheDocument();
+}
```
cerebrl, line 32: I do feel this is unnecessary. If we have a lot of these, it can slow down our CI and just generally cause delays, timeouts or other issues for very little or no actual benefit. If something is this simple, we can just not test it. I think it's okay for some things to not be tested.

cerebrl (commented):
Let's redirect our focus a bit on what we are testing here. Please apply the feedback across all related components as I didn't comment "same here" on each one. Let me know if you have any questions.

src/lib/components/primitives/label/label.stories.js
ryanbas21: For me, it renders `Input Label` on the storybook.

cerebrl: Yeah, that's because we don't pass through the text intended to be in the label in the `label.story.ts` component, which is incorrect.

src/lib/journey/stages/stages.stories.js
```diff
@@ -48,8 +49,43 @@ export const UsernamePassword = {
...
+  userEvent.type(password, 'Password123')
+
+  expect(canvas.getByLabelText('Password').value).toEqual('Password123')
+
+  await userEvent.tab();
+  await userEvent.tab();
+  const signin = canvas.getByRole('button', { name: 'Sign In' })
+  expect(signin).toHaveFocus();
+  await fireEvent.click(signin);
+  expect(UsernamePassword.args.submitForm).toHaveBeenCalled();
```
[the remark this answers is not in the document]
cerebrl: Please revisit this comment as well. I really would like to test our integration with the SDK callback APIs.

cerebrl (changes requested):
Let me know if you have any questions about the comments. Mostly I want to add assertions with the callback components integration of the SDK's callback APIs.

package-lock.json
```diff
@@ -2123,22 +2123,6 @@
...
-      ],
-      "dev": true,
-      "optional": true,
-      "os": [
-        "linux"
-      ],
-      "engines": {
-        "node": ">=12"
-      }
-    },
```
cerebrl, line 2141: I think the changes here to the package-lock.json are failing the deployment to Vercel. Can you take a look at it?

src/lib/journey/callbacks/password/password.stories.js
```diff
@@ -21,3 +23,23 @@ export const Base = {
...
+Interaction.args = { ...Base.args };
+Interaction.play = async ({ canvasElement }) => {
+  const canvas = within(canvasElement);
+  const passwordField = canvas.getByLabelText('Password')
+  expect(passwordField.value).toBe("")
+  await userEvent.tab();
+  expect(passwordField).toHaveFocus();
+
+  await userEvent.keyboard('password')
+  expect(passwordField.value).toBe("password")
```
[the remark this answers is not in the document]
ryanbas21: done.

src/lib/components/primitives/label/label.story.svelte
```diff
@@ -3,6 +3,10 @@
 
   export let key: string;
   export let classes: string;
+  export let inputLabel: string;
 </script>
 
-<Label {key} {classes}>Input Label</Label>
+<Label {key} {classes}
+  ><p>{inputLabel}</p>
+  <p /></Label
```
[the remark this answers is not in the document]
ryanbas21: I believe when I dove into the component, this used a `slot` and i don't think it made sense to just pass in the raw string, but rather an actual html element. Correct me if i'm wrong.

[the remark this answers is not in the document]
cerebrl: Yes, it uses a slot, but a label is almost always just a string. It _may_ contain inline elements, like `<strong>` or `<emphasize>`, but it rarely, if ever, contains block level elements like `<p>`. With that said, you also have a self-closing `<p />` tag for some reason. I'm not even sure that's valid HTML.

Either way, just passing in a plaintext statement is perfectly fine.

cerebrl (approved):

PR #28, opened 2022-10-27
ci: add-lighthouse-to-ci

ryanbas21:
adding lighthouse to ci adds different metrics that affect how a web app is understood by google as well as performance metrics. the better the lighthouse score, typically the more respect google gives to the application


cerebrl:
@ryanbas21 I have a question about Lighthouse: what is Lighthouse actually testing in regards to Accessibility? Is it actually crawling our site and evaluating the pages, or is it just statically evaluating from the initial render?

ryanbas21:
> 

https://developer.chrome.com/docs/lighthouse/accessibility/scoring/ 

based on the pages we tell it to run audits on, so based on the initial render. If we can edit the page state via url params then we could probably do something like that as well.

src/lib/journey/callbacks/password/base.svelte
```diff
@@ -89,8 +89,10 @@
     slot="input-button"
     type="button"
   >
-    <EyeIcon classes="tw_password-icon dark:tw_password-icon_dark" visible={isVisible}
-      ><T key="showPassword" /></EyeIcon
+    <EyeIcon
+      aria-label="show-password"
```
cerebrl, line 93: I'm assuming there's a reason you added this `aria-label`, but I would think this would be redundant since we are also passing a `title` attribute to the `EyeIcon`. All icons use the `<slot />` mechanism for this. You can see it here with the `EyeIcon`: https://github.com/cerebrl/forgerock-web-login-framework/blob/main/src/lib/components/icons/eye-icon.svelte#L19.

If the `title` attribute is not working well enough in your testing, then I'd like to change this underlying implementation, rather than sticking an `aria-label` on the element as doing so may end up duplicating what's read to the user.

tests/widget/modal/widget-modal.login.test.js
```diff
@@ -22,3 +22,54 @@ test('modal widget', async ({ page }) => {
...
+  expect(await dialog.isVisible()).toBeFalsy();
+
+  await loginButton.click();
+  // Add just a bit of a delay to ensure dialog responds
+  await page.waitForTimeout(1000);
+  expect(await dialog.isVisible()).toBeTruthy();
+  await page.goto('widget/modal');
+  const username = page.getByRole('textbox', { name: /User Name/ });
+  const password = page.getByRole('textbox', { name: /Password/ });
+  const eye = page.getByRole('button', { name: 'show-password' });
```
cerebrl, line 38: You should be able to find this button with `"Show Password"` and not need the addition of the `aria-label`. Here's a screenshot of the current app with the Accessibility panel open: https://share.getcloudapp.com/7KukDq6P. As you can see, the button should have an accessible name, but if it isn't working reliably, then let me know.

tests/widget/modal/widget-modal.login.test.js
```diff
@@ -22,3 +22,54 @@ test('modal widget', async ({ page }) => {
   expect(await fullName.innerText()).toBe('Full name: Demo User');
   expect(await email.innerText()).toBe('Email: demo@user.com');
 });
+
+test.only('modal widget accessibly', async ({ page }) => {
+  const loginButton = page.locator('button', { hasText: 'Open Login Modal' });
+  const dialog = page.locator('dialog');
+  expect(await dialog.isVisible()).toBeFalsy();
+
+  await loginButton.click();
```
cerebrl, line 31: I'd like these a11y-focused tests to not have any mouse related events, just as a convention.

tests/widget/modal/widget-modal.login.test.js
```diff
@@ -22,3 +22,54 @@ test('modal widget', async ({ page }) => {
...
+  const eye = page.getByRole('button', { name: 'show-password' });
+
+  const loginBtn = page.getByRole('button', { name: 'Sign in' });
+  await page.keyboard.press('Tab')
+  expect(username).toBeFocused();
+  await page.type(username, 'username01');
+
+  await page.keyboard.press('Tab')
+  expect(password).toBeFocused();
+  await page.type(password, 'password');
```
cerebrl, line 47: Can we split the username and password from each other (i.e. don't use a page node)? I'd like this a11y test to encounter a multi-step journey to test our focus management.

tests/widget/modal/widget-modal.login.test.js
```diff
@@ -22,3 +22,54 @@ test('modal widget', async ({ page }) => {
...
+
+  await page.keyboard.press('Tab')
+  expect(eye).toBeFocused();
+
+  await page.keyboard.press('Tab')
+  expect(loginBtn).toBeFocused();
+  await page.keyboard.press('Enter')
+
+  const error = page.getByRole('section')
+  const tryAgain = page.getByRole('button', { name: 'Please try again' });
```
cerebrl, line 57: I'm a bit confused by this. I would expect _this_ view to be rendered after a failed login attempt: https://share.getcloudapp.com/wbuQ2YdK. This view shouldn't have a button with "Please try again." The view it seems you're getting is an unrecoverable error, rather than just a login failure.

tests/widget/modal/widget-modal.login.test.js
```diff
@@ -22,3 +22,54 @@ test('modal widget', async ({ page }) => {
...
+  await page.type(password, 'password');
+
+  await page.keyboard.press('Tab')
+  expect(eye).toBeFocused();
+
+  await page.keyboard.press('Tab')
+  expect(loginBtn).toBeFocused();
+  await page.keyboard.press('Enter')
+
+  const error = page.getByRole('section')
```
cerebrl, line 56: I'd like to make sure each element we are testing for focus is also tested for their accessible text. If I understand this correctly, the above (as an example) `page.getByRole('textbox', { name: 'User Name' });` would fail if there was no "textbox" with accessible text of "User Name", right? I'd like us to do the same for errors and subsequent journey views.

tests/widget/modal/widget-modal.login.test.js
```diff
@@ -22,3 +22,54 @@ test('modal widget', async ({ page }) => {
...
+  await page.type(username, 'username01');
+
+  await page.keyboard.press('Tab')
+  expect(password).toBeFocused();
+  await page.type(password, 'password123!');
+
+  await page.keyboard.press('Tab')
+  await page.keyboard.press('Tab') // eye icon here
+  await page.keyboard.press('Enter')
+
```
cerebrl, line 74: We should probably test what happens after the authentication is successful and the widget closes. We shouldn't abandon our focus management even though we are done authenticating, you know?

tests/widget/modal/widget-modal.login.access.test.js
```diff
@@ -0,0 +1,61 @@
+import { expect, test } from '@playwright/test';
+
+test('modal widget accessibly', async ({ page }) => {
```
daveadams56, line 3: Typo - think this should be "modal widget accessibility"?

.github/workflows/preview-closed.yml
```diff
@@ -0,0 +1,34 @@
...
+        with:
+          AM_URL: ${{ secrets.AM_URL }}
+          USERNAME: ${{ secrets.AM_USERNAME }}
+          PASSWORD: ${{ secrets.AM_PASSWORD }}
+          REALM_PATH: ${{ secrets.AM_REALM }}
+          ORIGINS: ${{ steps.preview_urls.outputs.preview_urls }}
+          COOKIE_NAME: ${{ secrets.COOKIE_NAME }}
+          REDIRECTION_URIS: ${{ secrets.REDIRECTION_URIS }}
+          REMOVE: true
+          CORS_CONFIG_NAME: 'org-ui'
```
daveadams56, line 30: Are we creating a CORS config called 'org-ui', or expecting it to be there? I remember that at some point the default was no longer named 'org-ui'. Maybe it has been reverted now, but we might be better off creating our own for the previews, if we are not already doing so.

.github/workflows/ci.yml
```diff
@@ -7,8 +7,8 @@ jobs:
   pr:
     strategy:
       matrix:
-        node: [16.x] # tests both npm 7 & 8
-        os: [macos-latest, ubuntu-latest, windows-latest]
+        node: [16.x]
+        os: [macos-latest, ubuntu-latest] # windows-latest is commented out because it breaks on adapter-node
```
daveadams56, line 11: Do we have a JIRA to fix this? Or is it something we can't really control / work around?

daveadams56 (approved):
This is great to see. Do we have associated stories to improve the scores, or for now are we just implementing lighthouse? Just left a few comments and questions but otherwise happy.

svelte.config.js
```diff
@@ -1,12 +1,18 @@
...
 /** @type {import('@sveltejs/kit').Config} */
 const config = {
   extensions: ['.svelte', '.md'],
   kit: {
-    adapter: adapter(),
+    adapter: process.env.PREVIEW ? node() : auto(),
+  },
+  server: {
+    watch: {
+      usePolling: true
```
cerebrl, line 14: Can you share why you added this?

tests/widget/inline/widget-inline.login-2step.access.js
```diff
@@ -0,0 +1,29 @@
+test('2step login with inline accessiblity', async ({ page }) => {
```
cerebrl, line 1: Can you add `.test` to this filename to make it consistent with the others? Also, I'm not a huge fan of the use of "access" to be short for accessibility. Access already has a very strong meaning since we are in Access Management. Can we use "a11y" instead, both in code, strings as well as filenames?

tests/widget/modal/widget-modal.login-2step.access.test.js
```diff
@@ -0,0 +1,43 @@
+import { expect, test } from '@playwright/test';
+
+test('2step login with access', async ({ page }) => {
```
cerebrl, line 3: Here's the other instance of "access" that I think we should avoid. Either the full word "accessibility" or "a11y", I'd recommend.

cerebrl (approved):
Other than the few comments, I think this looks good. Please reply to the existing comments and adjust the naming, and this can be merged in.

svelte.config.js
ryanbas21: sveltekit uses chokidar under the hood to watch files and that doesn’t work in a docker container, so if we are to be able to develop in a docker container itself with okteto, to get HMR you need that polling config.

PR #29, opened 2022-11-09
feat(suspended-id): add support for suspended id

cerebrl:
## Summary

Adds support for the Email Suspend Node and "journey continuation" after the clicking of a "magic link".

## Details

- Support for `SuspendedTextOutputCallback` already existed
- Implement support for handling `suspendedId` from "magic link"
- Fix issue with `authIndexValue` being appended to `/authenticate` when sending `suspendedId`
- Fixing app and e2e test implementation that set a default tree on initialization (having a `tree` value on the base config makes it difficult to handle journey continuation)
- Added documentation for new APIs as well as tutorial

## Manual test

https://forgerock-web-login-framework-git-featsuspended-emai-d87688-cix.vercel.app/?journey=ResetPassword

Make sure to test with a user that has a valid email address you can access. Check your email and just copy the parameters portion of the link (without clicking on it). The paste the parameters replacing the existing parameters on the above URL. 

For example (don't actually use):

https://forgerock-web-login-framework-git-featsuspended-emai-d87688-cix.vercel.app/?realm=/alpha&suspendedId=Ii4yP7fSeaq2N4SZsdnoQ45JYqI


src/lib/config.ts
```diff
@@ -73,7 +73,7 @@ const configSchema = z
         z.literal('localStorage'),
       ])
       .optional(),
-    tree: z.string(),
+    tree: z.string().optional(),
```
cerebrl, line 76: I discovered that we need to discourage the setting of the `tree` value on the base config. It's best if they pass in a tree at actual call time.

src/lib/journey/journey.interfaces.ts
```diff
@@ -35,6 +35,7 @@ export interface CallbackMetadata {
 export interface JourneyStore extends Pick<Writable<JourneyStoreValue>, 'subscribe'> {
   next: (prevStep?: StepTypes, nextOptions?: StepOptions) => void;
   reset: () => void;
+  resume: (url: string, nextOptions?: StepOptions) => void;
```
cerebrl, line 38: Adding a `resume` function here that just wraps `next`, but allows the passing of the `url` argument.

src/lib/journey/journey.store.ts
```diff
@@ -54,10 +54,25 @@ export function initialize(initOptions?: StepOptions): JourneyStore {
     });
 
     try {
-      /**
-       * Initial attempt to retrieve next step
-       */
-      nextStep = await FRAuth.next(prevStep as FRStep, options);
+      if (resumeUrl) {
+        // If resuming an unknown journey remove the tree from the options
+        options.tree = undefined;
```
cerebrl, line 59: If `tree` exists, the SDK will append `authIndexType` and `authIndexValue` to the `/authenticate` call. If this value differs from the tree associated with the `suspendedId`, it will not work.

src/lib/journey/journey.store.ts
```diff
@@ -54,10 +54,25 @@ export function initialize(initOptions?: StepOptions): JourneyStore {
...
+
+        /**
+         * Attempt to resume journey
+         */
+        nextStep = await FRAuth.resume(resumeUrl, options);
+      } else if (prevStep) {
+        // If continuing on a tree remove it from the options
+        options.tree = undefined;
```
cerebrl, line 67: Also discovered in my testing of Transactional Authorization (SDK, not Widget) that journey continuation works best without `authIndex*` parameters. So, removing this if we have a previous step to prevent bugginess.

src/lib/journey/stages/stages.stories.js
```diff
@@ -80,80 +80,80 @@ RegistrationInteraction.play = async ({ canvasElement }) => {
   const news = canvas.getByRole('checkbox', { name: /news/ })
   const securityQuestion = canvas.getByLabelText('Select a security question')
   const securityAnswer = canvas.getByLabelText('Security Answer')
-  const toc = canvas.getByRole('checkbox', { name: 'Please accept our Terms and Conditions' });
+  const tocLink = canvas.getByRole('link', { name: 'View full Terms & Conditions' });
+  const toc = canvas.getByRole('checkbox', { name: 'Please accept our Terms & Conditions' });
```
cerebrl, line 84: Fixing a failed test because of the small change in the TOC component from a previous PR.

src/lib/journey/stages/stages.stories.js
```diff
@@ -80,80 +80,80 @@ RegistrationInteraction.play = async ({ canvasElement }) => {
...
   const emailCb = frRegistrationStep.getCallbacksOfType(CallbackType.StringAttributeInputCallback)[2];
   const specialOffersCb = frRegistrationStep.getCallbacksOfType(CallbackType.BooleanAttributeInputCallback)[0]
   const newsOffersCb = frRegistrationStep.getCallbacksOfType(CallbackType.BooleanAttributeInputCallback)[1]
-
   const securityQuestions = frRegistrationStep.getCallbacksOfType(CallbackType.KbaCreateCallback)[0];
-
   const tocCb = frRegistrationStep.getCallbacksOfType(CallbackType.TermsAndConditionsCallback)[0]
 
-  expect(username).toHaveFocus();
+  await expect(username).toHaveFocus();
```
cerebrl, line 96: My VS Code was barking about this not having an `await`, likely from my Storybook extension. We should add awaits to the rest of the tests as well over time.

src/routes/docs/contributing/testing/+page.md
```diff
@@ -12,8 +14,170 @@ We test components with Storybook's "Interaction" tests for our integration of S
 
 We test the fully running artifact with Playwright. Testing the widget is done against a running Svelte app that imports the built package of the Widget as if installed via NPM. Tests are in the root `tests` directory.
 
+### Manual tests
+
+These are end-to-end tests that cannot be automated and must be manually tested. This usually involves randomly generated codes that reside outside of the system, and are hard or impossible to mock.
+
+#### Please manually test the following
+
+1. Journey continuation feature (e.g. Email Suspend Node) requires manual testing due to the randomly generated code sent to email which cannot be automated at this time.
```
cerebrl, line 23: Noting that this will be a manual test, at least until we develop a mock server that could help automate this.

src/routes/docs/contributing/testing/+page.md
```diff
@@ -12,8 +14,170 @@ We test components with Storybook's "Interaction" tests for our integration of S
...
+
 ## Troubleshooting
 
+### Playwright Test Async Issue Explained
```
cerebrl, line 27: Adding the Playwright async issue explanation as it may help a future dev.

src/routes/docs/widget/full-api/+page.md
```diff
@@ -31,7 +31,7 @@ const widget = new Widget({
       realmPath: 'alpha', // OPTIONAL; Uses 'alpha', if not provided
       redirectUri: window.location.href, // OPTIONAL; falls back to `window.location.href`
       scope: 'openid email', // OPTIONAL; falls back to minimal 'openid email'
-      tree: 'Login', // OPTIONAL; falls back to default Login journey provided in ForgeRock
+      tree: 'Login', // NOT RECOMMENDED; OPTIONAL; falls back to default Login journey provided in ForgeRock
```
cerebrl, line 34: We might want to start changing how we approach setting the `tree` value on the base config.

src/routes/docs/widget/tutorials/+page.md
```diff
@@ -7,3 +11,49 @@ Below are tutorials for popular Web frameworks/libraries, like React, Angular an
 ### Angular with Angular CLI
 
 ### Modal with Vue and Vite
+
+### "Magic links" and journey continuation
```
cerebrl, line 15: I'm not totally sold on placing this doc here, but just wanted to put it somewhere for now.

src/lib/journey/stages/stages.stories.js
ryanbas21: yeah ive been ignoring these errors because i don't understand why a non promise i needs to be awaited?

src/lib/config.ts
ryanbas21: is there further reasoning as to why we should discourage this? All of our samples and docs kinda follow this pattern.

src/routes/(app)/+page.svelte
```diff
@@ -52,7 +54,17 @@
 
   // Use if not initializing journey in a "context module"
   onMount(async () => {
-    journeyStore.next();
+    // debugger;
```
ryanbas21, line 57: do we want this comment?

src/routes/(app)/+page.svelte
```diff
@@ -52,7 +54,17 @@
...
+
+    if (suspendedIdParam) {
+      // Resume journey
```
ryanbas21, line 60: non blocking but unneccessary comment. I think the api speaks for itself here.

src/routes/(app)/+page.svelte
```diff
@@ -52,7 +54,17 @@
...
+      journeyStore.resume(location.href);
+
+    } else {
+      journeyStore.next(undefined, {
```
ryanbas21, line 64: I understand why we are passing `undefined` here right now, but i must say the api is not super desirable this way. I wonder if we could do something about this. 

`journeyStore` could have a `setup` method or a `options` method where we can configure this value and then pass through the tree information?

An api like this is what i'm thinking

```js
else {
  journeyStore
    .configure({ tree: journeyParam || authIndexValue || undefined })
     .next();
     
 ```
 
 I'm wondering if with `typescript` we could then use a `generic` on `configure` or whatever we call it, so that the `step` can be inferred and passed through to subsequent method calls.
 
 This is all very rough in my head but i think we may be able to sketch out something if you understand where i'm going. Think the generics at different levels exercise we did the other week.

**Review by ryanbas21 — CHANGES_REQUESTED** [review]
left a few questions/comments/thoughts

File: src/lib/config.ts (hunk 1 above)
Comment (cerebrl) [reply src/lib/config.ts:76]: I'll write something up in the Widget's docs about this today. That way, we will have it documented for ourselves and others.

File: src/lib/journey/stages/stages.stories.js (hunk 6 above)
Comment (cerebrl) [reply src/lib/journey/stages/stages.stories.js:96]: Okay, after some digging, I found the reason: https://github.com/storybookjs/eslint-plugin-storybook/blob/main/docs/rules/await-interactions.md.

It looks like the original implementations of the Testing Library methods are wrapped by Storybook and to provide the full functionality of Storybook's Interactions, they recommend using `await`.

File: src/routes/(app)/+page.svelte (hunk 13 above)
Comment (cerebrl) [reply src/routes/(app)/+page.svelte:64]: Yeah, I get what you're saying. I too have been somewhat on the fence about this. As I've been working through this API design, I've been trying to find the right balance between introducing novel patterns and keeping some consistency with the underlying SDK. That way, those that may use the SDK is some projects, but this Login Framework in others (internal ForgeRockers is one such person), they don't have to flip back-n-forth between different `.next()` implementations.

Now, it's also worth mentioning that this is the Login App portion of code, and this API is not exposed to the Widget user. So, for now, this can be considered internal APIs. But, let's talk it out as I get your concern.

I'm also hesitant to add a `.configure()` method when we also have an `.initialize()` method that has to be run first, that accepts a configuration object. We also already have a `.configure()` method here: https://github.com/cerebrl/forgerock-web-login-framework/blob/6c3949cedcecc562312801b2789cf5c5e71e62aa/src/routes/(app)/%2Blayout.ts#L6.

How about this: What if I leverage the existing SDK convention of having a `.start()` method on this object? That would remove the need for the `undefined` step and keep us consistent with the SDK's methods. Does that sound doable?

**Review by ryanbas21 — APPROVED** [review]

File: src/routes/docs/contributing/faqs/+page.md [hunk 14]
```diff
@@ -145,3 +145,16 @@ This was always true for XUI and is still somewhat true for Platform Login. Cust
...
+### Question: Why do we not recommend setting a default `tree`?
+
+There are some edge cases when using journeys/trees in ways other than initially authenticating a user. A few examples are:
+
+1. Journey continuation with the Email Suspend Node
+2. Transactional Authorization
+
+In these examples, journeys are started _without_ explicitly declaring the `authIndexValue` (a reference to the journey) in the `/authenticate` request. Now, if we set a `tree` on the base config, the JavaScript SDK will, by default, add that as a value on the `authIndexValue` parameter. This can lead to conflicts between the journey the `suspendedId` or `advices` implicitly uses and the journey the `authIndexValue` is referencing.
+
+To fix this, we recommend removing the configuration of the tree on base configurations, and only pass the intended tree at the start of a journey. This also better aligns with the other SDKs as they only append the `authIndexType` and `authIndexValue` at the start of calling a journey, but not on the subsequent calls to the endpoint.
```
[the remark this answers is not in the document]
daveadams56: Actually, maybe we don't need this seeing as this is a contributing doc..

src/routes/docs/widget/tutorials/+page.md
```diff
@@ -7,3 +11,49 @@ Below are tutorials for popular Web frameworks/libraries, like React, Angular an
...
+
+Verifying a user's email address, building a "forgot password" flow, or using an email as a two-factor authentication method, ForgeRock's Email Suspend Node within journeys is the tool for the job. This results in the journey/tree being suspended until the user clicks a link (aka "magic link") found in their email. This link contains a randomly generated code that is used to continue the suspended journey.
+
+Here's how to take advantage of this feature:
+
+First, add the Email Suspend Node to the portion of the journey you'd like to suspend until the user continues the journey from the link found in their email.
```
daveadams56, line 21: I think you could save yourself a screenshot and provide more context here by linking to the Backstage docs for Suspended Authentication - there is an example tree there too:

https://backstage.forgerock.com/docs/am/7/authentication-guide/authn-suspended.html#authn-suspended

src/lib/journey/journey.store.ts
daveadams56: Actually having tested this change and observed the network requests, it seems that suspendedId is all that the authenticate endpoint needs to resume, so everything you said above makes sense 👍

src/lib/journey/journey.store.ts
daveadams56: Makes sense now I understand this a bit better

daveadams56 (approved):
This PR looks good to me. I've taken a look at the code and left a handful of non-blocking comments. I also manually tested this in a couple of browsers and tried non-happy paths e.g. invalid or already-used suspended ID values, and this performs well. Approved!

src/routes/docs/widget/tutorials/+page.md
cerebrl: Agreed. I'll be cleaning up the docs for the rest of this sprint, so I'll add this to the small task list I have locally. Since I have two approvals and a green build, I'll save our CI from having to run a full build just for a small change to the doc.

PR #31, opened 2022-11-16
update-infra

ryanbas21:
Ticket: https://bugster.forgerock.org/jira/browse/SDKS-2185

Update core dependencies, I moved e2e to the live deployment to use less CI resources and i was having an isuse with standing up the CI and e2e and this one was just easier. 

I had to implement a retry but I haven't fully investigated why the tests that get retried are flakey. I will address in a follow up. I can remove the live e2e if preferred, but i feel it provides a more realistic e2e experience since its on a deployment.


cerebrl:
@ryanbas21 Since this PR is now ready for review, can you provide the customary details and conventions to the title, description and add any comments to help focus our attention, if needed?

.github/workflows/ci.yml
```diff
@@ -7,7 +7,7 @@ jobs:
   pr:
     strategy:
       matrix:
-        node: [16.x]
+        node: [16.x, 18.x]
```
ryanbas21, line 10: added node 18 here

.github/workflows/ci.yml
```diff
@@ -20,13 +20,9 @@ jobs:
...
           cache: 'npm'
       - run: npm ci
-      - run: npm install -g @lhci/cli@0.8.x
       - run: npm run build
-      - run: npx playwright install  --with-deps
-      - name: E2E
-        run: npm run ci:e2e
       - name: Publish to Chromatic
-        if: ${{ matrix.os == 'ubuntu-latest' }}
+        if: ${{ matrix.os == 'ubuntu-latest' && matrix.node < 17 }}
```
ryanbas21, line 25: only want to publish to chromatic one time and chromatic isn't availble on node 18 from what i can tell.

.github/workflows/preview.yml
```diff
@@ -82,3 +82,21 @@ jobs:
           header: lighthouse
           message: |
             ${{ steps.lighthouse_result.outputs.comment }}
+  e2e:
```
ryanbas21, line 85: add e2e to the preview workflow so that we run e2e's after the deployment is finished.

.node-version
```diff
@@ -1 +1 @@
-v16.17.0
+v18.12.1
```
ryanbas21, line 1: I guess preference. maybe there is a way to provide multiple node versions, i just haven't found it.

package.json
```diff
@@ -28,80 +28,80 @@
...
-    "svelte2tsx": "^0.5.16",
-    "tailwindcss": "^3.1.8",
-    "tslib": "^2.3.1",
-    "typescript": "~4.7.4",
-    "vite": "^3.1.0",
-    "vitest": "^0.16.0"
+    "svelte2tsx": "^0.5.20",
+    "tailwindcss": "^3.2.4",
+    "tslib": "^2.4.1",
+    "typescript": "^4.9.3",
```
ryanbas21, line 95: !!!!

package.json
```diff
@@ -28,80 +28,80 @@
...
+    "vite": "^3.2.4",
+    "vitest": "^0.25.2"
   },
   "type": "module",
   "engines": {
-    "node": ">= 16.0.0 <17.0.0",
+    "node": ">16.0.0 < 17.0.0 || 18.12.1",
```
ryanbas21, line 101: because we can use either node version, i set it like so.

package.json
```diff
@@ -28,80 +28,80 @@
...
     "npm": ">=8.0.0"
```
ryanbas21, line 102: I may advise we bump to npm 9 but didn't want to force that change.

playwright.config.ts
```diff
@@ -1,20 +1,22 @@
 /** @type {import('@playwright/test').PlaywrightTestConfig} */
-import { devices } from '@playwright/test';
+import { devices, PlaywrightTestConfig } from '@playwright/test';
 
-const url = 'https://localhost:3000';
-const config = {
-  webServer: {
+const url = process.env.PLAYWRIGHT_TEST_BASE_URL || 'https://localhost:3000';
```
ryanbas21, line 4: if we have a base url, run on that url if not use localhost.

playwright.config.ts
```diff
@@ -1,20 +1,22 @@
 /** @type {import('@playwright/test').PlaywrightTestConfig} */
-import { devices } from '@playwright/test';
+import { devices, PlaywrightTestConfig } from '@playwright/test';
```
ryanbas21, line 2: I just wanted the autocompletion so i made the file a ts file and brought in the definition.

playwright.config.ts
```diff
@@ -1,20 +1,22 @@
...
 
-const url = 'https://localhost:3000';
-const config = {
-  webServer: {
+const url = process.env.PLAYWRIGHT_TEST_BASE_URL || 'https://localhost:3000';
+const config: PlaywrightTestConfig = {
+  webServer: process.env.CI ? null : {
```
ryanbas21, line 6: if we are in CI we don't have a web server to wait on.

playwright.config.ts
```diff
@@ -1,20 +1,22 @@
...
-    url,
+    reuseExistingServer: true
   },
   use: {
     headless: !!process.env.CI,
     baseURL: `${url}/e2e/`,
     ignoreHTTPSErrors: true,
+    trace: 'on'
   },
+  retries: process.env.CI ? 2 : 0,
```
ryanbas21, line 17: this is what i added for flakiness in ci. have to investigate and maybe we can remove. i think official playwright repo uses this though. 0 means 8 workers in this case or hte max when not on CI.

rollup.config.js
```diff
@@ -8,12 +8,11 @@ import json from '@rollup/plugin-json';
 import postcss from 'rollup-plugin-postcss';
 import resolve from '@rollup/plugin-node-resolve';
 import svelte from 'rollup-plugin-svelte';
-import { terser } from 'rollup-plugin-terser';
+import terser from '@rollup/plugin-terser';
```
ryanbas21, line 11: new terser plugin is in rollup repo.

src/hooks.server.ts
```diff
@@ -0,0 +1,7 @@
+import type { RequestEvent } from '@sveltejs/kit';
```
ryanbas21, line 1: svelte update has a difference between server / client in some way. so this file changed to be this.

src/routes/(app)/+layout.svelte
```diff
@@ -12,7 +12,7 @@
 
   <meta charset="utf-8" />
   <title>Login Application</title>
-  <link rel="icon" href="/favicon.ico" />
+  <link rel="icon" href="/favicon.png" />
```
ryanbas21, line 15: honestly this still throws but i have no idea how the favicon was being picked up to begin with.

src/routes/e2e/widget/inline/+page.svelte
```diff
@@ -2,7 +2,7 @@
   import { onMount } from 'svelte';
   import { page } from '$app/stores';
 
-  import Widget, { form, journey, user } from '$package/inline';
+  import Widget, { form, journey, user } from '$package/inline.svelte';
```
ryanbas21, line 5: another weird thing but i had to add this for stuff to work. it really didn't like this path

src/routes/e2e/widget/modal/+page.svelte
```diff
@@ -2,7 +2,7 @@
   import { onMount } from 'svelte';
   import { page } from '$app/stores';
 
-  import Widget, { modal, journey, user } from '$package/modal';
+  import Widget, { modal, journey, user } from '$package/modal.svelte';
```
ryanbas21, line 5: same as above

.github/workflows/ci.yml
ryanbas21: Yes but it also errors on 18 so i had to use just 16.

playwright.config.ts
cerebrl: The flakiness is due to not using the localhost server within Github actions?

src/routes/docs/+layout.svelte
```diff
@@ -1,6 +1,6 @@
 <script>
-  import TopNav from "./top-nav.svelte";
-  import SideNav from "./side-nav.svelte";
+  import TopNav from './top-nav.svelte';
```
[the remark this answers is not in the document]
cerebrl: Single quotes is correct for JS. Our formatter just uses double quotes for HTML attributes.

playwright.config.ts
ryanbas21: i think its actually more related to the test. the webpage does start up and is waiting. i think its test related.

src/routes/e2e/widget/inline/+page.svelte
cerebrl: Huh, interesting. This would definitely not be desired since this is the user facing package, and I wouldn't want our users to have to import something that looks like a Svelte component. Concerning ...

ryanbas21: the compiled file ends in `.svelte.js` so if we can remove the .svelte part we should be fine

playwright.config.ts
cerebrl: Huh, that's odd. I've actually noticed the tests being _more_ stable locally and on Github actions. So, this seems concerning that we've lost the progress.

src/routes/e2e/widget/inline/+page.svelte
cerebrl: Then let's definitely look into how to remove the `.svelte` portion of the file name.

playwright.config.ts
```diff
@@ -1,20 +1,22 @@
...
   forbidOnly: !!process.env.CI,
-  workers: process.env.CI ? 2 : 4,
+  workers: process.env.CI ? 1 : 0,
```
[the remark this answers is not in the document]
ryanbas21: this was wrong, it's `undefined` not 0. so i made that fix.

src/routes/e2e/widget/inline/+page.svelte
ryanbas21: from rollup docs:
```
This pattern will also be used for every file when setting the [output.preserveModules](https://rollupjs.org/guide/en/#outputpreservemodules) option. Note that in this case, [name] will include the relative path from the output root and possibly the original file extension if it was not one of .js, .jsx, .mjs, .cjs, .ts, .tsx, .mts, or .cts.
```

so i added a .replace on it. seemed to work, let me know if you don't like it though

daveadams56 (approved):
Reviewed with Ryan and offline, and tested the preview app. Approved.

PR #32, opened 2022-11-17
SDKS-2112 First pass at login widget strings

forgerock-chris:
Just some consistency and simplification suggestions.
Wasn't sure about the two errors - are these displayed to end users? Might have to simplify them if so, and offer a solution, even if it's just "Try again later."


cerebrl (approved):
Looks good to me. I'll wait for feedback from Giessel to merge.

PR #33, opened 2022-11-17
task(docs): add more information to docs

cerebrl:
## Summary

Just adding more content to our docs within the Contributing section. None of this content is final as I'm just trying to get words onto the the page as I think of them.

## Bug fix

There was a very small visual bug with the animated radio buttons not being centered, so there's a small fix for that.

### DIFF
--- a/src/routes/docs/widget/roadmap/+page.md
+++ b/src/routes/docs/widget/roadmap/+page.md
@@ -1,6 +1,6 @@
-## Future APIs (not implemented)
+# Future APIs (not implemented)
 
-### Currently **unsupported**
+## Currently **unsupported**
 
 1. WebAuthn
 2. Push Authentication
@@ -13,7 +13,7 @@
 9. Central Login
 10. SAML
 
-### Widget customization (future)
+## Widget customization (future)
 
 ```js
 new Widget({
@@ -28,7 +28,7 @@ new Widget({
 });
 ```
 
-### Additional modal events (future)
+## Additional modal events (future)
 
 ```js
 modal.onClose((event) => {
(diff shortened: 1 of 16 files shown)


themes/default/compositions.cjs
```diff
@@ -274,9 +274,11 @@ module.exports = (config, theme) => ({
   '.animated-radio': {
     position: 'relative',
     cursor: 'pointer',
-    fontSize: '1.5em',
-    height: '1em',
-    width: '1em',
+    // fontSize: '1.5em',
+    // replacing `1em` with `24px`
+    // using `em` results in misaligned circles due to half-pixel (e.g. 22.5px) grid alignment issue
```
cerebrl, line 279: When using `em`, the browser may calculate the width and height to be a non-integer value (e.g. 22.5 pixels). That can lead to misalignment as these items fall onto the pixel grid. Using a round, even number instead, like `24px` makes it easier to create a predictable alignment.

daveadams56 (approved):
Overall looks good - just a couple of very minor points if you get time.

themes/default/compositions.cjs
cerebrl: Good question. `fontSize` was the attribute driving the base value that informed the `em` unit value in `height` and `width`, which was the underlying issue. Switching to absolute values removed the need for declaring a `fontSize`.

src/routes/docs/widget/roadmap/+page.md
```diff
@@ -1,6 +1,6 @@
-## Future APIs (not implemented)
+# Future APIs (not implemented)
```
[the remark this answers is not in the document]
cerebrl: Good point. I'll make a note of this locally and include it in a different PR.

PR #34, opened 2022-11-25
feat(callbacks): add social login callbacks

daveadams56:
Jira: https://bugster.forgerock.org/jira/browse/SDKS-2113

## Summary

This PR adds support for `SelectIDPCallback` and `RedirectCallback` supporting social login flows with Apple as an identity provider.


src/lib/components/icons/apple-icon.svelte
```diff
@@ -0,0 +1,20 @@
...
+  />
+  <path
+    d="M11.182.008C11.148-.03 9.923.023 8.857 1.18c-1.066 1.156-.902 2.482-.878 2.516.024.034 1.52.087 2.475-1.258.955-1.345.762-2.391.728-2.43zm3.314 11.733c-.048-.096-2.325-1.234-2.113-3.422.212-2.189 1.675-2.789 1.698-2.854.023-.065-.597-.79-1.254-1.157a3.692 3.692 0 0 0-1.563-.434c-.108-.003-.483-.095-1.254.116-.508.139-1.653.589-1.968.607-.316.018-1.256-.522-2.267-.665-.647-.125-1.333.131-1.824.328-.49.196-1.422.754-2.074 2.237-.652 1.482-.311 3.83-.067 4.56.244.729.625 1.924 1.273 2.796.576.984 1.34 1.667 1.659 1.899.319.232 1.219.386 1.843.067.502-.308 1.408-.485 1.766-.472.357.013 1.061.154 1.782.539.571.197 1.111.115 1.652-.105.541-.221 1.324-1.059 2.238-2.758.347-.79.505-1.217.473-1.282z"
+  />
+  <title><slot /></title>
+</svg>
```
daveadams56, line 20: Taken from [Wikimedia](https://commons.wikimedia.org/wiki/File:Apple_logo_black.svg) - if anyone has a better source for the social icons, please let me know.

src/lib/journey/_utilities/metadata.utilities.ts
```diff
@@ -40,17 +46,39 @@ export function buildCallbackMetadata(
       isSelfSubmitting: isSelfSubmitting(callback),
       isUserInputRequired: requiresUserInput(callback),
       idx,
+      social: buildCallbackSocialMetadata(callback),
```
daveadams56, line 49: Slightly unsure about whether this is the best way to achieve this. The problem I am trying to solve is how to let the stage know that despite this being a self-submitting callback, we still need a "Next" button if the `Select IDP Node` has the `Local Authentication` option enabled.

src/lib/journey/_utilities/metadata.utilities.ts
```diff
@@ -77,12 +105,28 @@ export function isSelfSubmitting(callback: FRCallback) {
...
 /**
  * @function isStepSelfSubmittable -
  * @param {array} callbacks - CallbackMetadata
  * @returns
  */
 export function isStepSelfSubmittable(callbacks: CallbackMetadata[]) {
+  if (callbacks.filter((callback) => callback.social)) {
+    return true;
+  }
+
```
daveadams56, line 129: I found that the desired behaviour for the SelectIDPCallback is slightly different to other callbacks in terms of self-submission. An example of this is the fact that clicking one of the social provider buttons should submit the form regardless of the state of any of the other callbacks on the page - unlike other "Self Submitting" callbacks which generally need there to be no other "Unsubmittable" callbacks.

By the way when I say "desired" behaviour, I mean the expected user experience on such a page - I would expect to click "Continue With Apple" for example, and to skip anything else on the page (e.g. Username, Password).

src/lib/journey/callbacks/idp/idp.mock.ts
```diff
@@ -1,47 +0,0 @@
-export const docsExample = {
```
daveadams56, line 1: I removed this since this type of callback is what is returned for mobile devices, instead of web. For web apps the "Select IDP" callback is returned.

src/lib/journey/callbacks/redirect/redirect.svelte
```diff
@@ -0,0 +1,26 @@
...
+  export let stepMetadata: StepMetadata;
+  export let style: Style = {};
+
+  export let step: FRStep;
+
+  FRAuth.redirect(step);
+</script>
+
+<div class="tw_text-center">
+  <Spinner colorClass="tw_text-primary-light" layoutClasses="tw_h-24 tw_mb-6 tw_w-24" />
```
daveadams56, line 25: I made the decision for this callback to just show a spinner, and not take any text to display. I'm happy to be overruled but I'm not sure what we would show here.

src/lib/journey/callbacks/select-idp/select-idp.svelte
```diff
@@ -0,0 +1,66 @@
...
+  function setBtnValue(value: string) {
+    callback.setProvider(value);
+    callbackMetadata.isReadyForSubmission = true;
+    selfSubmitFunction && selfSubmitFunction();
+  }
+
+  // Assume that clicking "next" will indicate the user wants to use local authentication
+  if (callbackMetadata.social?.isLocalAuthenticationEnabled) {
+    callback.setProvider('localAuthentication');
+  }
```
daveadams56, line 40: As the above comment says, I think the cleanest / simplest way to show a "Local Authentication" option to the user is to by default set the provider to be "Local Authentication", and let the "Next" button on the page handle it. It means that we don't need to worry about there being another "submit" button on the page and that button displaying below all other callbacks, since "Next" already does this. Also, this is what Platform UI provides.

src/lib/journey/journey.interfaces.ts
```diff
@@ -31,6 +31,7 @@ export interface CallbackMetadata {
   isSelfSubmitting: boolean;
   isUserInputRequired: boolean;
   idx: number;
+  social?: Maybe<SocialMetadata>;
```
daveadams56, line 34: I figured that when defining metadata around the social aspect of a callback, it would be best for the callback metadata to use a Maybe so that a `SocialMetaData` being present would imply that the callback is a social one (handy to know in other parts of the app), and that we don't necessarily over-burden the metadata with another setting (i.e. every callback explicitly stating that it is / isn't a social callback, and having a mostly superfluous property in indicating local authentication). However I'm working on my TypeScript, so @cerebrl @ryanbas21 I would appreciate your view on how this could best be structured! 😄

src/lib/journey/stages/generic.svelte
```diff
@@ -112,12 +112,13 @@
...
       callbackMetadata={callbackMetadataArray[idx]}
       selfSubmitFunction={determineSubmission}
+      {step}
       stepMetadata={{ ...stepMetadata }}
       style={$style}
     />
   {/each}
 
-  {#if !stepMetadata.isStepSelfSubmittable}
+  {#if !stepMetadata.isStepSelfSubmittable || stepMetadata.isStepProvidingLocalAuth}
```
daveadams56, line 121: This is what I meant in my earlier comment about "_we still need a "Next" button if the Select IDP Node has the Local Authentication option enabled._". The Select IDP callbacks are self submittable, but otherwise don't play by the rules!

src/routes/(app)/+page.svelte
```diff
@@ -57,8 +58,9 @@
 
   // Use if not initializing journey in a "context module"
   onMount(async () => {
-    if (suspendedIdParam) {
+    if (suspendedIdParam || formPostEntryParam) {
       journeyStore.resume(location.href);
+      history.replaceState({}, '', '/');
```
daveadams56, line 63: I noticed that we are not sanitizing the URL after resuming the flow, whether from Suspended or Social login, so decided to look at how we can mutate the query parameters using Svelte Kit. My findings soon showed that this is not possible (or at least doesn't seem to be possible), so instead I invoked the browser API to rewrite the current URL. @cerebrl as a Svelte ninja there may well be a better way you know of to do this, so please share your thoughts!

daveadams56:
For the time being this is still a bit of a work in progress and needs some form of e2e testing, but I think there is enough here for dialog on how this has been implemented.

src/lib/journey/callbacks/redirect/redirect.svelte
cerebrl: We will need to have some text to ensure this is accessible for users of screen readers. We also want to make sure it's read to the user assertively, so we are sure it's been communicated. The ARIA Live attribute with `"assertive"` should do it: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions.

src/lib/journey/_utilities/metadata.utilities.ts
cerebrl: I think what you did on lines 156-159 of this file is reasonable. I can't think of a reason why this wouldn't work, and it's a similar pattern for `ConfirmationCallback`.

src/lib/journey/callbacks/redirect/redirect.svelte
cerebrl: I think text of "Redirecting to <top domain>"? Or, something like that.

cerebrl: For example: If the user clicks on Google as their Identity Provider, the UI would say the following:

```
Redirecting you to google.com.
```

You could extract the domain from the URL we get from the callback, so it would work quite universally. Just a thought.

src/lib/journey/_utilities/metadata.utilities.ts
cerebrl: Okay, I think I have a suggestion for handling this in a generic way. I've made some notes locally, but it's getting a bit late for me today, so I'll send them your way tomorrow morning.

The reason this is a unique case because this specific callback is dynamic, and the form structure is not just driven by the existence of the callback, but conditional data inside the callback. We need to be able to handle three types of use-cases:

1. Social IdPs and username-password form
2. Social IdPs and local auth button (no form)
3. Social IdPs only

I'm thinking a prop on the step metadata object of `isUserInputOptional`, which would help inform the stage component to know if it should render the "Next" button or not. This would satisfy the condition of both the step being "self submittable" and having user input that's optional.

To determine this, I think we should have custom functions that can be ran for determining this, but let me play around with this idea tomorrow.

src/lib/journey/callbacks/redirect/redirect.svelte
daveadams56: This makes sense - the domain text is easy enough, and I'll look into the ARIA stuff you mentioned. Thanks.

tests/widget/modal/widget-modal.login-social-redirect.test.js
```diff
@@ -0,0 +1,15 @@
...
+
+  const loginButton = page.locator('button', { hasText: 'Open Login Modal' });
+  await loginButton.click();
+
+  await page.fill('text="User Name"', 'demouser');
+
+  await page.locator('button', { hasText: 'Continue with Google' }).click();
+
+  await page.waitForNavigation();
+  expect(page.url()).toContain('accounts.google.com', 'redirect_uri');
```
daveadams56, line 14: When it comes to e2e testing for this I'm aware that some of it inherently needs to be manual since we are dealing with external services, accounts etc. However one bit that I thought could be auto tested was the redirect callback - to keep the test simple I've assumed that our test tenant will provide a Google social IDP and that we just use that to verify that the user is redirected to the Google accounts page as part of an OAuth flow. I originally looked at extracting the redirectURI from the AM response and verifying that we get sent there, however it felt like overkill in this test file and might bloat the tests, so instead went for this simpler test. Let me know what you think @cerebrl.

daveadams56: Also I concluded that I don't think we should test the "return" part of the flow since the logic to identify the query params lives in the developer's app and the JS SDK itself, rather than any widget functionality. I opted to not test this.

daveadams56: Anyway let me know if there are any obvious gaps in e2e that you find, and let's discuss. Thanks.

src/lib/journey/callbacks/redirect/redirect.svelte
```diff
@@ -0,0 +1,34 @@
...
+  import type { Style } from '$lib/style.store';
+  import type { Maybe } from '$lib/interfaces';
+
+  export let callback: RedirectCallback;
+  export let callbackMetadata: CallbackMetadata;
+  export let selfSubmitFunction: Maybe<SelfSubmitFunction> = null;
+  export let stepMetadata: StepMetadata;
+  export let style: Style = {};
+
+  export let step: FRStep;
```
daveadams56, line 20: Injecting the step into the component felt slightly wrong since we don't do this for any other callbacks, however I needed access to it for the SDK call to work `FRAuth.redirect() `expects the entire step rather than just the callback. If there's a nicer way to achieve this I'm open to ideas.

daveadams56:
A note about testing this change:

In order to test the widget changes, you should be able to go to the preview app and append`?journey=LoginWithSocial` to the URL. This should show the login page with **Google** and **Local Authentication** (Next), as IDP options since Google is configured on the test tenant. Clicking on Google should show the form with a Redirect callback, and you should be taken to accounts.google.com. The testing for the flow using the preview app ends there however, since the client configuration expects the app to be on `example.com`.

Additionally there are Storybook tests and e2e for the Redirect callback.

src/lib/journey/_utilities/metadata.utilities.ts
cerebrl: Okay, here's what I've got:

1. At the step level, we can iterate over the callback metadata, and add a property of `canForceUserInputOptionality`, or something, which has a boolean value, essentially saying this callback can force normally required user input callbacks to be optional
2. We can then add a new that step property, `isUserInputOptional`, which indicates that the step does indeed have user input type callbacks, but they are optional in this instance

Regarding the first item, I think we should provide a way to analyze callbacks by iterating through the step's callbacks and applying a generic function that could come from a user of the framework, so they could alter this behavior.

```js
// This eventually will be overridable by user of framework
const forceUserInputOptionalityCallbacks = {
  SelectIdPCallback: (callback) => {
    return !!callback
      .getProviders()
      .find((provider) => provider.provider === 'localAuthentication');
  },
};

// Our existing callback metadata builder with extra property to inform
// if any callbacks can override the presence of user input requirements
export function buildCallbackMetadata(
  step: WidgetStep,
  checkValidation: (callback: FRCallback) => boolean,
) {
  return step?.callbacks.map((callback, idx) => {
    return {
      // This one is tough to name, but it really just communicates the
      // ability of a special callback that can self-submit regardless
      // of the presence of user input callbacks
      canForceUserInputOptionality: canForceUserInfoOptionality(callback),
      isFirstInvalidInput: checkValidation(callback),
      isReadyForSubmission: isCbReadyByDefault(callback),
      isSelfSubmitting: isSelfSubmitting(callback),
      isUserInputRequired: requiresUserInput(callback),
      idx,
    };
  });
}

// Out existing step metadata builder with an extra property to inform
// if the step has a unique combination of self-submitting callback as
// well as user input type callbacks
export function buildStepMetadata(callbackMetadataArray) {
  return {
    isStepSelfSubmittable: isStepSelfSubmittable(callbackMetadataArray),
    isUserInputOptional: isUserInputOptional(callbackMetadataArray),
    numOfCallbacks: callbackMetadataArray.length,
    numOfSelfSubmittableCbs: callbackMetadataArray.filter((cb) => !!cb.isSelfSubmitting).length,
    numOfUserInputCbs: callbackMetadataArray.filter((cb) => !!cb.isUserInputRequired).length,
  };
}

export function forceUserInputOptionalityCallbacks(callback) {
  // See if a callback function exists within this collection
  const fn = forceUserInputOptionalityCbs[callback.getType()];

  // If there is a function, run it and it will return a boolean
  return fn && fn();
}

// Notice this function can take a user provided argument function to 
// override behavior (this doesn't have to be well defined)
export function isUserInputOptional(callbackMetadataArray, fn) {
  // default reducer function to check if both overriding callback exists
  // along with user input required callbacks
  const fallbackFn = (prev, curr) => {
    if (curr.canForceUserInfoOptionality && callbackMetadataArray.numOfUserInputCbs > 0) {
      prev = true;
    }
    return prev;
  };
  // Call reduce function with either fallback or user provided function
  return callbackMetadataArray.reduce(fn || fallbackFn, false);
}
```

Now that we have the above information, we can use this to generically affect the rendering of the form. So, if `stepMetadata.isUserInputOptional` is true, then we know we have to render the form. If it's false, then we make sure the step isn't self-submittable. If it isn't the we render the "Next" button.

```html
{#if stepMetadata.isUserInputOptional || !stepMetadata.isStepSelfSubmittable}
  <Button busy={loading} style="primary" type="submit" width="full">
    <T key="nextButton" />
  </Button>
{/if}
```

I hope this all makes sense. Also, please keep in mind the above is just pseudocode and not verified to work.

tests/widget/modal/widget-modal.login-social-redirect.test.js
```diff
@@ -0,0 +1,15 @@
...
+
+test.only('Modal widget with social callback redirect', async ({ page }) => {
+  await page.goto('widget/modal?journey=LoginWithSocial');
+
+  const loginButton = page.locator('button', { hasText: 'Open Login Modal' });
+  await loginButton.click();
+
+  await page.fill('text="User Name"', 'demouser');
+
+  await page.locator('button', { hasText: 'Continue with Google' }).click();
```
ryanbas21, line 11: I think we should prefer the `getByRole` api in this case.

```suggestion
  await page.getByRole('button', { name: 'Continue with Google' }).click();
```

src/lib/journey/journey.interfaces.ts
ryanbas21: Is there a reason why we don't just `extend` this interface and create a `SocialMetadata` inferface?

something like 

```
interface SocialMetaDataCallback extends CallbackMetadata {
 social: SocialMetaData
 }
```

src/lib/journey/_utilities/metadata.utilities.ts
```diff
@@ -1,11 +1,17 @@
...
+import {
+  CallbackType,
+  ConfirmationCallback,
+  SelectIdPCallback,
+  type FRCallback,
+} from '@forgerock/javascript-sdk';
 
 import type { CallbackMetadata, WidgetStep } from '$journey/journey.interfaces';
 
 const selfSubmittingCallbacks: string[] = [
```
ryanbas21, line 10: Can we make this `as const` so the type is a union of callback types? @cerebrl is this required to be mutable?

ryanbas21 (commented):
non blocking so  I will leave them as a comment. I need more time to digest what Justin posted about that solution still.

src/lib/components/icons/apple-icon.svelte
cerebrl: I haven't actually looked at it yet visually, but I'm sure it's fine. The most important thing to do is optimize the SVG for Web (there are a few great free tools on the Web, like https://jakearchibald.github.io/svgomg/). Often times, there's a bunch of junk or redundant nodes in the file, but just scanning the source here, it looks pretty clean.

src/lib/journey/_utilities/metadata.utilities.ts
cerebrl: Kind of. This will eventually be an exposed setting users of the framework can modify to control the behavior of the app or widget.

cerebrl: I guess I never made this very clear, so let me put my thoughts here:

My intention with these data structures and the logic used within the utility files is to keep most of the logic on how to render the form out of the components and avoid callbacks having to be aware of, or directly influence, other callbacks. 

My goal is to build up these metadata objects in a very generic fashion as to ensure high reusability and low specificity. Then, pass them as immutable objects into the components, so each component can make decisions about itself and render accordingly.

To configure the behavior, a dev can just add and remove callbacks from particular lists to alter the form's behavior. This avoids the devs from having to know the specifics or internals of callbacks. The only exception is the idea I commented here: https://github.com/cerebrl/forgerock-web-login-framework/pull/34/files#r1035288585.

The reason that one is different is because it's not just the presence of a callback, but the actual internal state of that callback and its relationship with other callbacks that drives the behavior. So, for that, we need for the dev to pass in a function, rather than a static value.

I hope that makes sense.

src/lib/journey/_utilities/metadata.utilities.ts
cerebrl: I kind of changed my mind on this, but it was not this specifically, but just the overall approach to this architectural challenge. Let me know if this comment makes sense and how it affects what you wrote: https://github.com/cerebrl/forgerock-web-login-framework/pull/34/files#r1035288585.

src/lib/journey/callbacks/select-idp/select-idp.mock.ts
```diff
@@ -1,3 +1,39 @@
...
+                buttonClass: '',
+                buttonCustomStyleHover:
+                  'background-color: #000000; color: #ffffff; border-color: #000000;',
+                buttonDisplayName: 'Apple',
+                iconClass: 'fa-apple',
+                iconFontColor: 'white',
+                iconBackground: '#000000',
+              },
+            },
+            { provider: 'localAuthentication' },
```
cerebrl, line 26: Let's ensure we test the 3 different "modes" for social login:

1. Social IdPs only (buttons for Social IdPs only)
2. Social IdPs and local auth, but no other callbacks present (buttons for social and button for local, no form)
3. Social IdPs and local auth with username-password callbacks present (buttons for social only, button for local is not present, and form for immediately entering username and password)

We should have these different states tested as different stages in Storybook as well as our end-to-end tests. Let me know if you need any additional clarification.

src/lib/journey/_utilities/metadata.utilities.ts
ryanbas21: my thought was really, how can we create an array that is as specific as we can be without breaking the "generic" desire of the code.

string[] feels too unsafe. We can perhaps as const this now with the intention of allowing a developer to `Pick` or `Omit` values from this type? 

Maybe i'm reaching a bit though?

src/lib/journey/callbacks/select-idp/select-idp.svelte
cerebrl: I think this makes sense. I'll continue thinking on this, but I like how this keeps callbacks independent.

src/lib/journey/journey.interfaces.ts
cerebrl: The suggestion I made earlier should replace this high specificity property with a more generic one: `canForceUserInputOptionality`.

src/lib/journey/stages/generic.svelte
cerebrl: As seen in the earlier comment, I'd like to keep use-case specific code out of the generic, high-level components like this one. Hopefully, my suggestion will help here.

src/routes/(app)/+page.svelte
cerebrl: I think the SvelteKit API you're looking for is the `goto` method found in `$app/navigation`:

```
goto('/', { replaceState: true });
```

Docs: https://kit.svelte.dev/docs/modules#$app-navigation-goto

src/lib/journey/_utilities/metadata.utilities.ts
cerebrl: Fair point. As long as we have a way for developers to easily manage the types in case they modify them, I'm good.

tests/widget/modal/widget-modal.login-social-redirect.test.js
cerebrl: Yeah, I think this is good. Let's make a note that the full flow needs to be manually tested. I think I have a section in the docs regarding use-cases that need to be manually tested. Can you add this to it?

tests/widget/modal/widget-modal.login-social-redirect.test.js
```diff
@@ -0,0 +1,15 @@
+import { expect, test } from '@playwright/test';
+
+test.only('Modal widget with social callback redirect', async ({ page }) => {
+  await page.goto('widget/modal?journey=LoginWithSocial');
+
+  const loginButton = page.locator('button', { hasText: 'Open Login Modal' });
+  await loginButton.click();
+
+  await page.fill('text="User Name"', 'demouser');
```
cerebrl, line 9: This is probably not needed for the test-case of using Google, but it brings up a good point. We have logic in the framework that manages self-submission and user-input optionality. We should test these permutations. 

So, having a test that locally logs a user in, with the presence of a social login option would be good. It would be good to test the absence of the local auth button. Then, we should have a test for local auth, but not form, which would show a local auth button.

daveadams56: Yes I noticed this and thought that was a mistake. However yes I think you're right, testing the combinations of social auth, local auth, social auth without any local auth etc makes sense. I'll build this out.

tests/widget/modal/widget-modal.login-social-redirect.test.js
daveadams56: Yes, I will do. Thanks for reminding me.

src/routes/(app)/+page.svelte
daveadams56: Ah I must have missed that. Perfect, thanks.

src/lib/journey/journey.interfaces.ts
daveadams56: Nice, I'm reviewing your suggestions now - I think I had a similar idea in an earlier commit before I went down the social metadata path - the idea of "forcibly self submitting" i.e. ignoring all other user input and self submitting. I'll work through it. Thanks.

src/lib/components/icons/apple-icon.svelte
daveadams56: Thanks for the recommendation of the tool - I'll bookmark that.

src/lib/journey/callbacks/redirect/redirect.svelte
cerebrl: Okay, this is a good callout. Let's definitely avoid this as it breaks the Rule of Least Privilege. Passing a callback component the whole step is giving it too much rope. Let's move the actual redirection into the stage component as it manages the data at the step level. At the end of the day, the rendering of the UI and the redirecting of the browser are separate things, and don't need to be coupled.

src/lib/journey/stages/generic.svelte
daveadams56: Just a thought as I'm reworking the changes based on your other comments:

Each stage has its own version of the login form and therefore "next" button, so any logic like the above will need to be duplicated between them. To keep things consistent in case for example we have a social flow as part of a registration journey, I'll need to make the same changes to the "registration" and "username-password" stage files. Similarly for the "redirect" functionality - each stage will have to implement this.

Is there a way that we can avoid duplication of code like this and therefore minimise risk of mistakes, in the eventuality that the number of stages grows?

cerebrl: There's no need to worry about the other stages as they are just examples of a named stage of known callbacks. Only the `generic` stage needs to handle anything thrown at it.

daveadams56: Ok makes sense. Thanks.

src/lib/journey/callbacks/redirect/redirect.svelte
daveadams56: This makes more sense. I've moved the redirect into the stage and a utility function. Let me know what you think.

src/lib/journey/_utilities/step.utilities.ts
```diff
@@ -55,6 +56,19 @@ export function initCheckValidation() {
...
+ * @function shouldRedirectFromStep -
+ * @returns {boolean}
+ */
+export function shouldRedirectFromStep(step: FRStep) {
+  try {
+    step.getCallbackOfType(CallbackType.RedirectCallback);
+    return true;
+  } catch (e) {
+    return false;
+  }
```
cerebrl, line 69: Rather than doing a `try-catch` with `getCallbackOfType` (which throws when the cb doesn't exist), I'd use `getCallbacksOfType` (which returns an array regardless of cb existence) and check the length of the array for what value to return.

src/lib/journey/callbacks/redirect/redirect.svelte
cerebrl: Nice, I think this works.

src/lib/journey/_utilities/step.utilities.ts
daveadams56: I see what you're saying, but I found getCallbacksOfType threw an error when there were no redirect callbacks. I'll try that again to be sure though.

cerebrl: Huh, it shouldn't, but if it does, we should make sure there isn't a regression. Here's the source for this method: https://github.com/ForgeRock/forgerock-javascript-sdk/blob/develop/packages/javascript-sdk/src/fr-auth/fr-step.ts#L61.

src/lib/journey/_utilities/metadata.utilities.ts
daveadams56: Sorry it's unclear whether you're both requesting a change here - what's your conclusion?

src/lib/journey/_utilities/metadata.utilities.ts
daveadams56: Ok this now makes sense to me - I have a working version reworked to your ideas and I like it. I need to look through what I wrote with a fresh mind but we're most of the way there I think. Thanks for putting so much thought and detail into your response! I'll push it probably tomorrow.

src/lib/journey/_utilities/step.utilities.ts
daveadams56: I will do more testing on it tomorrow, but it seems to throw an error - for example the above refactored to this:

```
if (step.getCallbacksOfType(CallbackType.RedirectCallback)) return true;
return false;
```

.. throws the following error in the console:

```
Uncaught (in promise) Error: Expected 1 callback of type "RedirectCallback", but found 0
    getCallbackOfType fr-step.ts:50
```

which doesn't make sense to me since it seems to be `getCallbackOfType()` that is failing despite calling `getCallbacksOfType`?

tests/widget/modal/widget-modal.login-social-redirect.test.js
```diff
@@ -0,0 +1,40 @@
...
+  await page.goto('widget/modal?journey=LoginWithSocialAndLocalAuth');
+
+  const loginButton = page.locator('button', { hasText: 'Open Login Modal' });
+  await loginButton.click();
+
+  await page.waitForNavigation();
+  expect(page.url()).toContain('accounts.google.com', 'redirect_uri');
+});
+
+test.only('Modal widget with social callback local authentication', async ({ page }) => {
```
daveadams56, line 25: @cerebrl what is your view on multiple test cases in the same file? I grouped them together as they are very similar variations on the same thing - the scenarios around local authentication. I'm happy to split them out into different files if you prefer.

ryanbas21:
@daveadams56 i couldn't verify this since the journey is on your tenant. i pushed up changes to get CI to run and remove the only from the PR.

src/lib/journey/_utilities/metadata.utilities.ts
ryanbas21: It's not a blocker to this PR. I just was suggesting adding the following:

```typescript
const selfSubmittingCallbacks = [
  CallbackType.ConfirmationCallback,
  CallbackType.DeviceProfileCallback,
  CallbackType.PollingWaitCallback,
  CallbackType.SelectIdPCallback,
] as const;

export type SelfSubmittingCallbacks = typeof selfSubmittingCallbacks[number]

const userInputCallbacks = [
  CallbackType.BooleanAttributeInputCallback,
  CallbackType.ChoiceCallback,
  CallbackType.ConfirmationCallback,
  CallbackType.KbaCreateCallback,
  CallbackType.NameCallback,
  CallbackType.NumberAttributeInputCallback,
  CallbackType.PasswordCallback,
  CallbackType.ReCaptchaCallback,
  CallbackType.SelectIdPCallback,
  CallbackType.StringAttributeInputCallback,
  CallbackType.TermsAndConditionsCallback,
  CallbackType.ValidatedCreatePasswordCallback,
  CallbackType.ValidatedCreateUsernameCallback,
] as const;

export type UserInputCallbacks = typeof userInputCallbacks[number]

```

This would be a little more restrictive from a type perspective. we would probably need to `as` cast the FRCallback because it isn't a generic type and getType() returns a string but this a bit more safe and makes sure the values aren't mutable at the type level so instead of "String" we would have a unionion of hte callbacks.

src/lib/journey/callbacks/redirect/redirect.svelte
```diff
@@ -0,0 +1,30 @@
...
+  export let callback: RedirectCallback;
+  export let callbackMetadata: CallbackMetadata;
+  export let selfSubmitFunction: Maybe<SelfSubmitFunction> = null;
+  export let stepMetadata: StepMetadata;
+  export let style: Style = {};
+
+  let message: string;
+
+  $: {
+    message = `Redirecting you to ${new URL(callback.getRedirectUrl()).hostname}.`;
```
daveadams56, line 23: Reminder to myself to localize this string

src/lib/journey/callbacks/select-idp/select-idp.svelte
```diff
@@ -0,0 +1,71 @@
...
+</script>
+
+<Grid num={idps.length}>
+  {#each idps as idp}
+    <Button type="button" width="auto" onClick={() => setBtnValue(idp.value)}
+      ><AppleIcon
+        classes="tw_inline-block tw_fill-current tw_text-secondary-dark dark:tw_text-secondary-light"
+      />
+      Continue with {idp.text}
```
daveadams56, line 64: Reminder to myself to localize this string

src/lib/journey/callbacks/select-idp/select-idp.mock.ts
daveadams56: I found late last week that mode "2" in this list is a bit of a grey area, in that the expected behaviour based on platform UI and implied by the docs, is that if local auth is enabled but no other callbacks are present, then there is no button for local and the local auth option is ignored. I have therefore implemented the widget, storybook and e2e this way and intend to raise a docs JIRA to make it more obvious that for local auth to work, there need to be other callbacks in the page node. @ryanbas21 @cerebrl tagging you for visibility in case you want to challenge this.

cerebrl: I'd like to challenge this, but let's not let it block the PR. When I originally worked on Social Login, the team (Bela at the time included) agreed to have a button for local authentication, which just allowed the journey designer to route that selection to a login form of the choosing. Again, we don't have to do it now, but I think this is needed for 1.0.

daveadams56: Ryan and agreed that for this story just following what Platform UI does is sufficient. I am happy for it to be included in a future story as it took me by surprise to find that platform UI doesn't render it - it seems illogical to have an outcome on the node that you can't get to by building your tree a certain way. Having 'local auth' route to a login form makes sense to me. I'll put it in another story. Thanks @cerebrl.

src/lib/journey/callbacks/select-idp/select-idp.svelte
```diff
@@ -0,0 +1,71 @@
...
+      text: option.uiConfig.buttonDisplayName,
+    }));
+  }
+</script>
+
+<Grid num={idps.length}>
+  {#each idps as idp}
+    <Button type="button" width="auto" onClick={() => setBtnValue(idp.value)}
+      ><AppleIcon
+        classes="tw_inline-block tw_fill-current tw_text-secondary-dark dark:tw_text-secondary-light"
```
daveadams56, line 62: In the demo to the team the look and feel of the Sign In With Apple button was raised, given that Apple have UX guidelines for this button: https://developer.apple.com/design/human-interface-guidelines/technologies/sign-in-with-apple/

There was also an unresolved discussion about whether we should build the callback to meet these requirements, or consume the uiconfig from AM. @ryanbas21 and I decided that that for this story we will treat the styling of the button as a non-blocking nice-to-have item and address the UX for all of the social buttons in the following story where other IDPs are added.

src/lib/journey/callbacks/select-idp/select-idp.mock.ts
daveadams56: Created [SDKS-2250](https://bugster.forgerock.org/jira/browse/SDKS-2250) for this.

src/lib/journey/_utilities/step.utilities.ts
daveadams56: I'll investigate further separate to this PR - @cerebrl I don't think this is blocking?

src/lib/journey/callbacks/select-idp/select-idp.svelte
daveadams56: UX of the social buttons has been added to the following social story where we will add other providers.

src/lib/journey/_utilities/step.utilities.ts
cerebrl: Let's revisit this in the future, but it's non-blocking at this point.

cerebrl (approved):

PR #35, opened 2022-12-06
feat(journey): add configurable journeys

cerebrl:


src/lib/_utilities/i18n.utilities.ts
```diff
@@ -108,8 +108,20 @@ export function interpolate(
...
+     * To preserve original behavior in addition to this one exception,
+     * return `undefined` for all other cases.
+     */
+    onTagAttr: function (tag, name, value) {
+      if (tag == 'a' && name == 'href') {
+        if (value.substr(0, 1) === '?') {
+          return `${name}="${value}"`;
+        }
+      }
+    }
```
cerebrl, line 123: When the `href` of the anchors started with a `?` for query parameters, they were getting stripped out of the text. This prevents that.

src/lib/journey/config.store.ts
```diff
@@ -0,0 +1,72 @@
...
+    match: ['#/service/Registration', '?journey=Registration'],
+  },
+};
+
+/**
+ * TODO: My intention is to type the parameter with `z.infer<typeof partialJourneysSchema>`,
+ * but when I do, the `map` does not like it. I tried a few things to fix it with no avail.
+ * I just threw an `any` on it for the time being.
+ */
+export function initialize(customJourneys?: any) {
```
cerebrl, line 52: This allows us to have configurable journeys within the app/widget. We fallback to the default journeys seen above if nothing is passed in.

For the life of me, I could not get the types right, and I ran out of time. If someone knows the magic spell, feel free to help me :)

src/lib/journey/journey.store.ts
```diff
@@ -3,13 +3,61 @@ import type { StepOptions } from '@forgerock/javascript-sdk/lib/auth/interfaces'
...
 import { htmlDecode } from '$journey/_utilities/decode.utilities';
-import type { JourneyStore, JourneyStoreValue, StepTypes } from './journey.interfaces';
+import type { JourneyStore, JourneyStoreValue, StackStore, StepTypes } from './journey.interfaces';
 import { interpolate } from '$lib/_utilities/i18n.utilities';
 import {
   authIdTimeoutErrorCode,
   shouldPopulateWithPreviousCallbacks,
 } from './_utilities/step.utilities';
 
+function initializeStack(initOptions?: StepOptions) {
```
cerebrl, line 13: This is the "stack" to provide the ability to "go back" to a previous journey. Since we can't rely on URL driven state (we don't control the URL), we need an internal history.

src/lib/journey/journey.store.ts
```diff
@@ -228,11 +277,25 @@ export function initialize(initOptions?: StepOptions): JourneyStore {
...
+    const currentJourney = updatedStack[updatedStack.length - 1];
+    await start(currentJourney);
+  }
+
+  async function push(newOptions: StepOptions) {
+    reset();
+    await stack.push(newOptions);
+    await start(newOptions);
+  }
+
```
cerebrl, line 292: This adds the stack's methods to the same `journey` store that is already being used. This prevents the "things" we need to pass around to a minimum.

src/lib/journey/journey.svelte
```diff
@@ -27,6 +29,10 @@
   afterUpdate(() => {
     alertNeedsFocus = !$journeyStore.successful;
   });
+
+  $: {
+    console.log($stack);
+  }
```
cerebrl, line 35: Oops, I'll remove this tomorrow.

src/lib/journey/journey.svelte
```diff
@@ -37,11 +43,19 @@
...
+        message: $journeyStore.error?.message || '',
+        status: $journeyStore.error?.code ? 'error' : 'ok',
+        submit: submitForm,
+      }}
+      journey={{
+        loading: $journeyStore.loading,
+        pop: journeyStore.pop,
+        push: journeyStore.push,
+        stack,
+      }}
```
cerebrl, line 58: I wanted to take this opportunity to start cleaning up and organizing the interfaces between the components. I think we can make it even better, but I needed to draw a line somewhere :)

src/lib/journey/stages/_utilities/stage.utilities.ts
```diff
@@ -0,0 +1,38 @@
+import { get } from 'svelte/store';
+
+import { configuredJourneys } from '$journey/config.store';
+
+import type { StageJourneyObject } from '$journey/journey.interfaces';
+
+export function captureLinks(linkWrapper: HTMLElement, currentJourney: StageJourneyObject) {
```
cerebrl, line 7: This was an important function to provide for OOTB behavior. The default journeys from ID Cloud are not named with a stage and therefor go through the generic stage component. The links that come from the Page Node's "Page Description" to forgot password and such are exactly that, links. So, I need to capture them, and internally handle the rendering of the selected journey, rather than navigate to the link's default `href` value (e.g. `#/service/ResetPassword`).

Just to reiterate, from the Widget's perspective, we can't rely on navigation since it's installed in a foreign app and making any assumptions about the parent app's route handling would be dangerous.

src/lib/journey/stages/_utilities/stage.utilities.ts
```diff
@@ -0,0 +1,38 @@
...
+        return item.match.find((string) => {
+          const href = anchor.getAttribute('href');
+          return href === string;
+        });
+      });
+
+      if (match) {
+        event.preventDefault();
+
+        const isPreviousJourney = stack.find((item) => {
```
cerebrl, line 26: Technically, this value would be an object, not a boolean, so I'll likely wrap this in a `!!()` to coerce the value to a boolean.

src/lib/journey/stages/generic.svelte
```diff
@@ -68,12 +69,13 @@
...
 
+  onMount(() => captureLinks(linkWrapper, journey));
+
   $: {
-    console.log(formNeedsFocus);
     checkValidation = initCheckValidation();
     callbackMetadataArray = buildCallbackMetadata(step, checkValidation);
     stepMetadata = buildStepMetadata(callbackMetadataArray);
-    failureMessageKey = convertStringToKey(failureMessage);
+    formMessageKey = convertStringToKey(form?.message);
```
cerebrl, line 78: I think we should move this into the `journey.store` to clean up the surface area for the stage component even more. I _really_ wanted to do it, but I restrained myself.

src/lib/journey/stages/generic.svelte
```diff
@@ -118,8 +120,10 @@
   {/each}
 
   {#if !stepMetadata.isStepSelfSubmittable}
-    <Button busy={loading} style="primary" type="submit" width="full">
+    <Button busy={journey?.loading} style="primary" type="submit" width="full">
       <T key="nextButton" />
     </Button>
   {/if}
+
+  <BackTo {journey} />
```
cerebrl, line 128: This is where the "back to" button is rendered when the user jumps from one journey to another.

src/lib/journey/stages/generic.svelte
```diff
@@ -84,12 +86,12 @@
...
   onSubmitWhenValid={submitFormWrapper}
 >
-  {#if displayIcon}
+  {#if form?.icon}
     <div class="tw_flex tw_justify-center">
       <ShieldIcon classes="tw_text-gray-400 tw_fill-current" size="72px" />
     </div>
   {/if}
-  <header id={formHeaderId}>
+  <header bind:this={linkWrapper} id={formHeaderId}>
```
cerebrl, line 94: This `linkWrapper` is what captures the specific link behavior and handles them internally if they match one of the configured journeys.

src/lib/journey/stages/generic.svelte
```diff
@@ -68,12 +69,13 @@
     }
   });
 
+  onMount(() => captureLinks(linkWrapper, journey));
```
cerebrl, line 72: This ensures the link wrapper and the capture links function bind after the component mounts. Since these links are added via HTML strings from an external source, we can't handle them natively in Svelte. So, we have to handle this is native JS.

src/routes/(app)/+layout.ts
```diff
@@ -1,4 +1,4 @@
-import configure from '$lib/config';
+import configure, { initialize as initializeJourneys } from '$lib/sdk.config';
```
cerebrl, line 1: This is an error, so I'll need to fix this tomorrow.

src/lib/journey/config.store.ts
```diff
@@ -0,0 +1,72 @@
...
+export let configuredJourneys: Readable<z.infer<typeof journeyItemSchema>[]>;
+
+/**
+ * Change `any` here to something that actually works
+ */
+const defaultJourneys: any = {
+  forgotPassword: {
+    journey: 'ResetPassword',
+    key: 'forgotPassword',
+    match: ['#/service/ResetPassword'],
```
cerebrl, line 28: These are the strings that we test for when a link is clicked. If an href matches one of these strings, then we know to internally handle the link and prevent default behavior. All other links are ignored and allowed through.

src/lib/journey/config.store.ts
```diff
@@ -0,0 +1,72 @@
...
+  login: journeyItemSchema,
+  register: journeyItemSchema,
+});
+
+export const partialJourneysSchema = journeysSchema.partial();
+
+export let configuredJourneys: Readable<z.infer<typeof journeyItemSchema>[]>;
+
+/**
+ * Change `any` here to something that actually works
```
ryanbas21, line 22: I took a whack at this cause I was bored tonight. I made some small changes, but if we better unify the keys of the object to the `key` property wthin the nested object and journey we can write a neat type.

```
type UncapitalizeObjectKeys<T> = {
  [key in keyof T as Uncapitalize<key & string>]: T[key] extends Object ? UncapitalizeObjectKeys<T[key]> : T[key]
}
type Journey = 'ResetPassword' | 'ForgottenUsername' | 'Registration' | 'Login'
type JourneyKeys = keyof UncapitalizeObjectKeys<{
    [K in Journey]: K
}>

type Key = Lowercase<Journey>
type Match = `#/service/${Journey}`

type Setup = { journey?: Journey, key: JourneyKeys, match: string[] }

const defaultJourneys = {
  resetPassword: {
    journey: 'ResetPassword',
    key: 'resetPassword',
    match: ['#/service/ResetPassword'],
  },
  forgottenUsername: {
    journey: 'ForgottenUsername',
    key: 'forgottenUsername',
    match: ['#/service/ForgottenUsername'],
  },
  login: {
    journey: undefined,
    key: 'login',
    match: ['#/service/Login', '?journey'],
  },
  registration: {
    journey: 'Registration',
    key: 'registration',
    match: ['#/service/Registration', '?journey=Registration'],
  },
} satisfies  Record<JourneyKeys, Setup>;
```

https://www.typescriptlang.org/play?ssl=35&ssc=41&pln=1&pc=1#code/C4TwDgpgBAqgdgYwIZgJbCQG1QLwgeQCMArCBYAaQhAGcAeAFQD4oBeKAbwCgooBtANbUoqOFCEgA9gDMoDKEhqxEKdFlwQ6EqADIoNYACdRAcyYBdAFxzB1c1AgAPYBDgATJUVLkoAfmXIaBjYeF5klNT0DLYg5izW0RLmXAC+XKCQUABSkgCuhnDC7ADkAEoQNBDAAAqKNADukoZuxVAAPlDFAGJNJpLALnAwlQVIALYQrR1lECaoBoZIwKiScFOdADKSc2vp4NA5+YUgVLRs4tQyAarBGmHkp-TcvLx8FCJihwV21hSpTFw9plTuctvUIIZkJU6F9jgCMtAALJLBAAC3OAAMAMQAehGADdUAgIDiACQcWHUFIYwEIqAAZSquTA5w4UGIeW+IF81kpIAANBcQLzOcdHoKxijUdYFqY+PY0lwEKsDFA3BBpEhcphgHylOxnlBDBUqrUaA0mm5rIbeByjtRrDNKjU6o1msV+TwXhJHcbnWaLe7PS8oJLgGjrHxiriCUSSeV-a7LcVzMGoCk09Jev1BsMIXBxhBrV7baKHZ0eoY+gNXHnRhMPSWhY6s1Wc7WRgWG2neGGI-xo3iIYTiTjK9Xc53Cym0xmvZhtqJiyG7VzrLl3BrRBA3D3m50FztGyG+9KBzHh3GcVsj4Lir5V8cZ1657xjXMFksVnBly9H+WZg-Iwv1WY9vQA995mA5ZQL3U9I0HWNR3KIDFhgtY7wfMsQFYFCoLQ79n14OcUn0L8aGkVAKl4cplWaGFsPFBkmTAJgAG4gA

ryanbas21: I think if we have something like ts-toolbelt we can make this work easier / better . I think.

ryanbas21: I also think we may be able to improve the type of `match` but i didn't wanna go down that road.

src/lib/journey/config.store.ts
ryanbas21: i'll have to pull this down and play with zod in this case

src/lib/journey/config.store.ts
daveadams56: Not sure I understand the meaning of the `match` property here - is this intended to be a URL path that users can also use to specify a given tree e.g. `https://sdkapp.example.com:8443/#/service/Registration`?

daveadams56: Oh I think I get this now - the 'match' is the value we check against for the Page Node "Description"? Does this mean that these links won't appear unless there is a match in the description? Is there another way to enable them?

src/lib/journey/journey.store.ts
daveadams56: Should this internal history persist between page reloads? It doesn't and I don't think it should, but want to check the thinking behind this.

src/lib/locale.store.ts
```diff
@@ -6,6 +6,8 @@ import fallback from '$locales/us/en/index.json';
 export const stringsSchema = z
   .object({
     alreadyHaveAnAccount: z.string(),
+    backToDefault: z.string(),
```
daveadams56, line 9: Maybe a silly question but why do we need backToDefault and backToLogin? How are they different?

daveadams56 (commented):
I've made an initial pass including a basic test of the functionality locally, but I'll need to spend a bit more time on this tomorrow. Just a few questions about how this works so far but I am liking the experience this provides!

src/lib/journey/journey.store.ts
cerebrl: This is a really good question and one I've been pondering for a while now. When using the modal form factor, I'd say it's pretty non-standard to save the internal state of what's inside the modal throughout a refresh. But, if one is using the inline form factor, maybe?

If they're using the inline form factor, I'd recommend that the devs render the links within the scope of their own app, which can correspond much better with navigations and preserving back button behavior, leaving just the form itself to the Widget. But, there's a hundred ways to build an app.

With that said, I don't think at this moment we should persist the state, but let's collect feedback after our initial release. If we should persist the state, we can have the discussion then. They way it's current built, I don't think adding this after the fact would be difficult.

Let me know what you all think.

src/lib/journey/config.store.ts
daveadams56: Ok I think I understand this better now. Thanks for explaining it.

src/locales/ca/en/index.json
```diff
@@ -1,6 +1,8 @@
 {
-  "alreadyHaveAnAccount": "Already have an account? <a href='/'>Sign in here!</a>",
-  "dontHaveAnAccount": "Don’t have an account? <a href='/register'>Sign up here!</a>",
+  "alreadyHaveAnAccount": "Already have an account? <a href='?journey'>Sign in here!</a>",
```
daveadams56, line 2: Similar to my below comment - the `Already have an account? Sign in here!` link doesn't seem to do anything, with just the URL updated.

src/locales/ca/en/index.json
```diff
@@ -1,6 +1,8 @@
...
+  "backToDefault": "Back to Sign In",
+  "backToLogin": "Back to Sign In",
+  "dontHaveAnAccount": "Don’t have an account? <a href='?journey=Registration'>Sign up here!</a>",
```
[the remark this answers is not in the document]
daveadams56: My guess would be that `onMount()` in the test app isn't being called if the URL is just changed here instead of a proper page redirect, so the journey store isn't being initialized with the tree from the `journeyParam` - just my initial thought but maybe there's something I'm doing wrong? Although looking at the `match` array for the Registration journey, it now seems like we're expecting to capture that link navigation and turn it into an init of the journey store with that journey instead of following the link, but that isn't working.

daveadams56 (commented):
I've focused my efforts on testing this locally and it's mostly good, but I found a few things that don't quite seem right so I've left comments. I'll keep going, wanted to share these now.

src/lib/journey/stages/_utilities/stage.utilities.ts
daveadams56: I agree with the way you've done it here. We'll need to make sure the documentation on this behaviour is quite explicit though, as I think this could get quite confusing to a developer otherwise!

src/lib/locale.store.ts
daveadams56: Ok that makes sense now - thanks!

src/locales/ca/en/index.json
cerebrl: This should be fixed now. Thanks for catching it.

src/locales/ca/en/index.json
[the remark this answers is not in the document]
cerebrl: There was some logic in the linkWrapper that broke. It should now be fixed.

tests/widget/inline/widget-inline.register.test.js
```diff
@@ -7,7 +7,7 @@ test('Inline registration widget', async ({ page }) => {
   await page.fill('text=Username', uuid());
   await page.fill('text=First Name', 'Demo');
   await page.fill('text=Last Name', 'User');
-  await page.fill('text=Email Address', 'demo@user.com');
+  await page.fill('text=Email Address', 'test@auto.com');
```
cerebrl, line 10: Changing the email used for auto-generated users to more easily manage the deletion of them afterwards. Plus, when using "forgot password" or "forgot username" if more than one user has the email you enter, the journey just fails with a 401.

src/lib/journey/config.store.ts
ryanbas21: I did it with typing `Match` a bit, again small changes are in here but it was fun. let me know if this is possible with these changes.

```typescript
import { readable, type Readable } from 'svelte/store';
import { z } from 'zod';

export const journeyItemSchema = z.object({
  journey: z.string(),
  key: z.string(),
  match: z.string().array(),
});
export const journeysSchema = z.object({
  default: journeyItemSchema,
  forgotPassword: journeyItemSchema,
  forgotUsername: journeyItemSchema,
  login: journeyItemSchema,
  register: journeyItemSchema,
});
export const partialJourneysSchema = journeysSchema.partial();
type UncapitalizeObjectKeys<T> = {
  [key in keyof T as Uncapitalize<key & string>]: T[key] extends Record<string, unknown> ? UncapitalizeObjectKeys<T[key]> : T[key]
}
type Journey = 'ResetPassword' | 'ForgottenUsername' | 'Registration' | 'Login'
type JourneyKeys = keyof UncapitalizeObjectKeys<{
  [K in Journey]: K
}>

type Key = keyof typeof defaultJourneys;

type Match<route extends string, B extends string = ''> = `#/${route}/${Journey}` | `?${B}${string}`

type Setup = { journey?: Journey, key: JourneyKeys, match: ReadonlyArray<Match<"service", "journey">>; }

const defaultJourneys = {
  resetPassword: {
    journey: 'ResetPassword',
    key: 'resetPassword',
    match: ['#/service/ResetPassword'] as const,
  },
  forgottenUsername: {
    journey: 'ForgottenUsername',
    key: 'forgottenUsername',
    match: ['#/service/ForgottenUsername'] as const,
  },
  login: {
    journey: undefined,
    key: 'login',
    match: ['#/service/Login', '?journey'] as const,
  },
  registration: {
    journey: 'Registration',
    key: 'registration',
    match: ['#/service/Registration', '?journey=Registration'] as const,
  },
} satisfies  Record<JourneyKeys, Setup>;

type PartialJourneyKeys = keyof z.infer<typeof partialJourneysSchema>
export function initialize(customJourneys: z.infer<typeof partialJourneysSchema>) {
  if (customJourneys) {
    const arr = Object.keys(customJourneys) as PartialJourneyKeys[]
    return readable(
      arr.map((key) => customJourneys[key]),
    );
  } else {
    const arr = Object.keys(defaultJourneys) as Key[];
    return readable(
      arr.map(
        (key) => {
          return defaultJourneys[key];
        },
      ),
    );
  }
}
export let configuredJourneys: Readable<z.infer<typeof journeyItemSchema>[]
```

src/lib/journey/journey.store.ts
```diff
@@ -3,13 +3,61 @@ import type { StepOptions } from '@forgerock/javascript-sdk/lib/auth/interfaces'
...
+            state = current.slice(0, -1);
+          } else {
+            state = current;
+          }
+          resolve([...state]);
+          return state;
+        });
+      });
+    },
+    push: async (options?: StepOptions): Promise<StepOptions[]> => {
```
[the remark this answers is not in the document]
cerebrl: No Svelte specific anything in here, really, other than the core store APIs. Are you just wanting to remove the unnecessary braces?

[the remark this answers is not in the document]
cerebrl: Oh, you're also wanting to remove the `new Promise`? I need to look closer at the `update` function that comes from the Svelte store. I think it's asynchronous, which is why I'm using the Promise. But, I'll see if I can make sure.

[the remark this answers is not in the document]
ryanbas21: it just seems redundant since its async already?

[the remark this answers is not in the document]
cerebrl: Okay, that's right! The `update` function doesn't return anything, regardless of it being sync or async. Here's the `update` method's type:

```
const update: (this: void, updater: Updater<StepOptions[]>) => void
```

All we can do is return the value to the callback function _within_ the `update`, but not return a value from `update` itself. If that makes any sense.

src/lib/journey/config.store.ts
```diff
@@ -0,0 +1,66 @@
+import { readable, type Readable } from 'svelte/store';
+import { z } from 'zod';
+
+export const journeyConfigItemSchema = z.object({
+  journey: z.string().optional(),
+  match: z.string().regex(/^(#\/service|\?journey)/, { message: 'HREF string must start with `?journey` or `#/service`' }).array(),
```
cerebrl, line 6: This gives us really strong value validation at the runtime, but not perfect on the TypeScript side. I think this is good enough for now, but we can revisit later.

CC @ryanbas21

src/lib/journey/journey.store.ts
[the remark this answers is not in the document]
cerebrl: BTW, by async, I don't mean `async-await`, I just mean from the generic since of asynchronous code as it uses the callback pattern and update doesn't have an actual return value we can control.

src/lib/journey/stages/username-password.svelte
```diff
@@ -83,12 +91,33 @@
...
-    <Button busy={loading} style="primary" type="submit" width="full">
+    <Button busy={journey?.loading} style="primary" type="submit" width="full">
       <T key="loginButton" />
     </Button>
   {/if}
 
+  <p class=" tw_my-4 tw_text-base tw_text-center tw_text-link-dark dark:tw_text-link-light">
+    <button
+      on:click|preventDefault={() => {
+        journey.push({ tree: 'ResetPassword' });
```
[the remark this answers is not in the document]
daveadams56: I appreciate too that the styling as links is up to the developer given this is in a stage file - which is one of the reasons I think this is not a priority.

daveadams56 (approved):
I have re-reviewed this, tested locally and run e2e and storybook. This looks brilliant, I like the experience you're providing here. There is one question I have around the expectation on seeing a link, but this is not a blocker to this PR. Approved!

src/lib/journey/stages/username-password.svelte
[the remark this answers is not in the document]
cerebrl: I think these are valid concerns, and it's why I brought it up in the demo last Friday. I definitely think there's some improvements here on this front, but I'd really like to see how this is used in the wild. This feature will be pretty easy to adjust according to feedback as it's very configurable.

PR #37, opened 2022-12-08
SDKS 2111/improve local and ci

ryanbas21:
- feat: add-husky
- feat: pre-hooks
- feat(commitizen-and-semantic-release): adds semantic release to CI and commitizen to commit messages
- refactor(src/lib/constants.ts): extends the type of import.meta


.husky/pre-push
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+CI=true npm run ci
+npm run test:e2e
```
cerebrl, line 5: Is this like an `if` statement? It runs the the first command if CI is `true` or the second command if it's `false`? This syntax is a bit mind-bending if it's intended.

.releaserc
```diff
@@ -0,0 +1,25 @@
+{
+    "branches": [
+        { "name": "main" },
+        "+([0-9])?(.{+([0-9]),x}).x",
+        { "name": "beta", "channel": "beta", "prerelease": true },
+        { "name": "alpha", "channel": "alpha", "prerelease": true },
+        { "name": "next", "channel": "next" }
```
cerebrl, line 7: Is `next` not prerelease?

cerebrl (commented):
Just a few questions.

.husky/pre-push
ryanbas21: No, these are two seperate commands, like you would run in your shell. if CI is false the test command is in watch mode automatically i think, thats why i put that there

.releaserc
ryanbas21: https://semantic-release.gitbook.io/semantic-release/recipes/release-workflow/distribution-channels

I followed this kinda flow through it. seems like next is not considered prerelease.

.husky/pre-push
cerebrl: Oh, got it. `npm run ci` is exactly the same as `npm run test`. Thanks.

.releaserc
cerebrl: Huh, interesting. Fair enough.

cerebrl (approved):

.releaserc
```diff
@@ -0,0 +1,28 @@
...
+      ],
+      "plugins": [
+        "@semantic-release/commit-analyzer",
+        "@semantic-release/release-notes-generator",
+        "@semantic-release/changelog",
+         ["@semantic-release/exec", {
```
ryanbas21, line 13: This should exec the script to add the latest release version to the package.json before npm publish

PR #38, opened 2022-12-08
ci: chromatic-on-non-draft-only

ryanbas21:
we only want to run chromatic on a pr which is not a draft


cerebrl (approved):
This looked like it worked, so I'm good with this. I also see that it's in your other PR. So, we can merge this in or close it and merge it in with your other work.

PR #53, opened 2022-12-12
feat(callbacks): add social login providers

daveadams56:
Jira: https://bugster.forgerock.org/jira/browse/SDKS-2215

## Summary

This PR adds support for Facebook and Google identity providers in the social login flows added in [PR #34 ](https://github.com/cerebrl/forgerock-web-login-framework/pull/34).


src/lib/components/icons/apple-icon-dark.svelte
```diff
@@ -8,11 +8,8 @@
   height={size}
   width={size}
   class={classes}
-  viewBox="0 0 24 24"
+  viewBox="0 0 24 20"
```
daveadams56, line 11: I had to play around with the properties for the social icon SVGs to get them to align consistently. Again I took them from Wikimedia, but if we know of a better place to get them I'm all ears!

src/lib/journey/_utilities/step.utilities.ts
```diff
@@ -61,12 +61,7 @@ export function initCheckValidation() {
  * @returns {boolean}
  */
 export function shouldRedirectFromStep(step: FRStep) {
-  try {
-    step.getCallbackOfType(CallbackType.RedirectCallback);
-    return true;
-  } catch (e) {
-    return false;
-  }
+  return step.getCallbacksOfType(CallbackType.RedirectCallback).length > 0;
```
daveadams56, line 64: It's really strange how using `getCallbacksOfType` now works for me - in the previous PR I was getting weird errors when using this so I used `getCallbackOfType` in a try catch. Anyway, it seems to be working how I would have originally expected it to now, which is nicer.

src/lib/journey/callbacks/select-idp/select-idp.stories.js
```diff
@@ -1,15 +1,21 @@
...
 
-const social = new FRStep(socialStep);
-const socialLocalAuthNoForm = new FRStep(socialStepLocalAuthNoForm);
-const socialLocalAuthWithFormStep = new FRStep(socialStepLocalAuthWithForm);
+const multipleProvidersLocalAuthForm = new FRStep(multipleProvidersLocalAuthFormStep);
+const multipleProvidersLocalAuthNoForm = new FRStep(multipleProvidersLocalAuthNoFormStep);
+const multipleProvidersNoLocalAuth = new FRStep(multipleProvidersNoLocalAuthStep);
+const singleProviderLocalAuthForm = new FRStep(singleProviderLocalAuthFormStep);
+const singleProviderLocalAuthNoForm = new FRStep(singleProviderLocalAuthNoFormStep);
+const singleProviderNoLocalAuth = new FRStep(singleProviderNoLocalAuthStep);
```
daveadams56, line 18: I felt it was beneficial to test the scenarios around local auth with single or multiple IDPs now that we are supporting more than one. If there are any I've left out let me know.

src/lib/components/icons/apple-icon-light.svelte
```diff
@@ -0,0 +1,18 @@
...
+  width={size}
+  class={classes}
+  viewBox="0 0 20 20"
+>
+  <path
+    fill="#fff"
+    d="M11.182.008C11.148-.03 9.923.023 8.857 1.18c-1.066 1.156-.902 2.482-.878 2.516.024.034 1.52.087 2.475-1.258.955-1.345.762-2.391.728-2.43zm3.314 11.733c-.048-.096-2.325-1.234-2.113-3.422.212-2.189 1.675-2.789 1.698-2.854.023-.065-.597-.79-1.254-1.157a3.692 3.692 0 0 0-1.563-.434c-.108-.003-.483-.095-1.254.116-.508.139-1.653.589-1.968.607-.316.018-1.256-.522-2.267-.665-.647-.125-1.333.131-1.824.328-.49.196-1.422.754-2.074 2.237-.652 1.482-.311 3.83-.067 4.56.244.729.625 1.924 1.273 2.796.576.984 1.34 1.667 1.659 1.899.319.232 1.219.386 1.843.067.502-.308 1.408-.485 1.766-.472.357.013 1.061.154 1.782.539.571.197 1.111.115 1.652-.105.541-.221 1.324-1.059 2.238-2.758.347-.79.505-1.217.473-1.282z"
+  />
+  <title><slot /></title>
+</svg>
```
daveadams56, line 18: I added a "light" and "dark" version of the Apple logo given it's the only one that needs to change depending on which mode we're in.

src/lib/journey/callbacks/select-idp/select-idp.svelte
```diff
@@ -53,20 +57,39 @@
...
-      />
-      <T key="signInWith" />
+{#each idps as idp}
+  <Grid num={1}>
+    <Button type="button" width="auto" onClick={() => setBtnValue(idp.value)}>
+      {#if idp.text.toUpperCase().includes('APPLE')}
+        {#if prefersDarkTheme}
+          <AppleIconLight
+            classes="tw_inline-block tw_fill-current tw_text-secondary-dark dark:tw_text-secondary-light"
+          />
```
daveadams56, line 72: @cerebrl I'd like your opinion on this section, for a couple of reasons.

Firstly, we're currently treating all of the social IDP provider buttons effectively the same from a styling perspective. In terms of button styles within the widget project we don't have many choices - e.g. light or dark, outline or filled seem to be our options. There are guidelines from each provider on how the button should be styled, worded and placed depending on what is around it. I plan to adjust the styling of each button to meet these requirements - my approach will likely be to create classes specific to each provider and pass them in to the Button component here. If you have any thoughts on how else you'd like this to be done (or want to object to doing this at all), it would be good to discuss sooner rather than later.

Secondly, as I've implemented this here we're hard coding support for just these three IDPs, relying on the IDP name matching and making assumptions about the icon and styling to use. I think we should have an "else" option giving a generic icon and style if we don't recognise the IDP. I also think this could be made configurable, for example if someone wants to pass in configuration for a new IDP or a different logo. I'm just not sure that we should make it configurable, given these are the three supported providers by the SDK. Additionally if we style according to guidelines, I'm not sure if it is worth us making it super flexible as this may be something no one wants to do.

Anyway, please let me know what you think on the above points and we can discuss further. Thanks.

src/locales/us/en/index.json
```diff
@@ -44,7 +45,6 @@
   "securityQuestions": "Security question(s)",
   "securityQuestionsPrompt": "Provide custom security question(s) and answer(s):",
   "showPassword": "Show password",
-  "signInWith": "Sign in with",
```
daveadams56, line 47: I replaced the default button text "Sign in with", with "Continue with" since it seems to be more applicable in scenarios other than logging in, e.g. registration. However further to my comment a bit further up about branding of buttons for different IDPs, we may need to show "Sign in with Apple", "Continue with Facebook", etc. This may get a bit messy. Not sure what we can do about this but it feels like we should be consistent in our defaults ("Continue with.." would be my preference) - and allow developers to override in the localization files.

tests/widget/modal/widget-modal.login-social-redirect.test.js
```diff
@@ -6,7 +6,7 @@ test('Modal widget with social callback redirect', async ({ page }) => {
   const loginButton = page.getByRole('button', { name: 'Open Login Modal' });
   await loginButton.click();
 
-  const button = await page.locator('button', { hasText: 'Sign in with Google' });
+  const button = await page.locator('button', { hasText: 'Continue with Google' });
```
daveadams56, line 9: I just updated the test here for the new default text for the social IDP button - whilst I was in the area though I had a think about e2e and whether we need additional use cases for this story. I am leaning towards there being enough coverage as it is - but please let me know if you agree / disagree with this, and what you feel any lack of coverage would be.

src/lib/components/icons/apple-icon-dark.svelte
daveadams56: All I can get from Facebook is a set of official PNGs, but no SVGs. I think Google is similar but need to check along with Apple. Seeing as we are referencing other brands now in this project, would we prefer official PNGs, or publicly sourced (wikimedia), SVGs as defaults? I think official PNGs would probably be best, but didn't want to work against the patterns in the rest of the app.

cerebrl: I'd say the Bootstrap icons are likely the safest, most consistent source of icons: https://icons.getbootstrap.com. Other notable are https://www.flaticon.com and, if you super desperate, https://worldvectorlogo.com.

src/lib/components/icons/apple-icon-light.svelte
cerebrl: Is the shape of the icon actually different? Or, are you just changing the color? I'd personally say all of the icons need to have the appropriate color for the light and dark theme. Are the other icons multi-color? Is that why Apple is different as it is a single color?

src/locales/us/en/index.json
cerebrl: Tough question. I think "Continue with" is the best choice. I looked through the guidelines of all three and I don't think they are super rigid on the text preceding the company name. This make sense as it would have to accommodate localization. Not every language is going to literally translate to "Sign in with" or whatever. I think they just want it to be descriptive of the action that is triggered by the button.

tests/widget/modal/widget-modal.login-social-redirect.test.js
cerebrl: I'm good with this being the only one.

src/lib/components/icons/apple-icon-light.svelte
daveadams56: Yes, we are just changing the colour depending on whether it is included within a light or dark button. Given that the Apple icon is either black or light grey, it felt right to include both and select the right one based on the preferred mode. Facebook and Google however seem to work on both light and dark backgrounds, given they are multi-color.

src/lib/journey/callbacks/select-idp/select-idp.svelte
daveadams56: Understood, I'm implementing something that will allow us to apply bgColor and textColor to buttons - also looking at additional properties that we might use if we ever decide to allow the uiConfig returned with the social callback to dictate style.

src/lib/components/icons/apple-icon-light.svelte
cerebrl: Okay, got it. When looking at this as an isolated component, I'd recommend we don't duplicate two components for the sake of colors. We should pass assign the color via CSS in the way all the icons are assigned colors: https://github.com/cerebrl/forgerock-web-login-framework/blob/main/themes/default/primitives.cjs#L34. Just make sure to remove to assign `none` to the `fill` attribute: https://github.com/cerebrl/forgerock-web-login-framework/blob/main/src/lib/components/icons/eye-icon.svelte#L15.

That way, you can have a light and dark variant without having to duplicate code. I'll continue this thread on a different comment as I'd like to jump into how we know if something needs to be light or dark.

daveadams56: I'm looking at bootstrap SVGs now and using the button text color to inform the color of the icon - if I can get this to work then provided plain color icons are within guidelines then I think I can get rid of the dark / light apple logo.

src/lib/journey/callbacks/select-idp/select-idp.svelte
```diff
@@ -53,20 +57,39 @@
       value: option.provider,
       text: option.uiConfig.buttonDisplayName,
     }));
+
+    prefersDarkTheme = window.matchMedia('(prefers-color-scheme: dark)').matches;
```
cerebrl, line 61: Since this is a widget that will be embedded within a foreign app, we should avoid making assumptions that the light or dark theme is chosen purely on the theme setting. Light or dark may be chosen manually by the user (maybe the parent app has a theme switch or profile setting), or the light or dark theme may be completely disabled by the dev and hard coded to light or dark.

Either way, unfortunately, this method is not doable for us. Tailwind is the library that takes care of this feature for us. We use the parent class "mode" to detect if the dev as assigned the dark class on the widget's parent element and Tailwind reacts accordingly. That means we need to lean on Tailwind's special classes `dark:tw_my-component_dark` and have all theme related mechanics in the CSS and outside of the JS.

Let me know if that doesn't make sense. Here's where we assign that class to the body element: https://github.com/cerebrl/forgerock-web-login-framework/blob/main/src/app.html#L10. Yes, we do it according to the system setting, but we can't know if the dev will do the same. Here's the Tailwind config setting: https://github.com/cerebrl/forgerock-web-login-framework/blob/main/tailwind.config.cjs#L3.

src/lib/journey/callbacks/select-idp/select-idp.svelte
```diff
@@ -53,20 +57,39 @@
...
-      ><AppleIcon
-        classes="tw_inline-block tw_fill-current tw_text-secondary-dark dark:tw_text-secondary-light"
-      />
-      <T key="signInWith" />
+{#each idps as idp}
+  <Grid num={1}>
+    <Button type="button" width="auto" onClick={() => setBtnValue(idp.value)}>
+      {#if idp.text.toUpperCase().includes('APPLE')}
+        {#if prefersDarkTheme}
+          <AppleIconLight
```
cerebrl, line 70: If we lean on Tailwind and CSS only, it will allow us to use just use one Apple icon component and reduce the amount of Svelte code we have to bundle with the widget.

src/lib/journey/callbacks/select-idp/select-idp.svelte
daveadams56: Ok, I see your point. I'm just trying to find a way to make the tailwind `dark:` class stuff work with custom button styles. I think I might have to define some new classes for this so I can assign a class for `light` and `dark` mode for each social provider button.

daveadams56: Making the buttons work for light mode according to design guidelines is easy using bootstrap SVGs (and therefore a single file for each provider):

Screenshot 2022-12-14 at 15 48 40

daveadams56: However dark mode complicates things

src/lib/components/icons/apple-icon-dark.svelte
daveadams56: I went with Bootstrap SVGs in the end.

src/lib/components/icons/apple-icon-light.svelte
daveadams56: We now have just the one Apple icon which takes the color of the text. Not sure what the difference is for the `fill` attribute - there is no visibile difference between `currentColor` which is set now, and `none` which you suggested. Does this need to be changed?

src/lib/journey/callbacks/select-idp/select-idp.svelte
daveadams56: As we discussed on slack, we are now styling the buttons using classes passed from the SelectIdp component. The result is that we can take advantage of the tailwind detection of light and dark mode.

src/lib/journey/callbacks/select-idp/select-idp.svelte
daveadams56: Done.

cerebrl:
I reviewed the code, and it looks great. Unfortunately, I can't test the full feature with our usual Vercel deployment, so I'm have to pull this down locally and test it. I did find a few visual issues captured here: https://share.getcloudapp.com/4gu4J7gx.

The Google icon is a bit out of proportion, slightly misaligned and quite close to the text. The horizontal rule isn't as subtle as it is in light mode, so that should be adjusted. It looks like you're missing some of the classes on the `<hr />` element, seen here: https://github.com/cerebrl/forgerock-web-login-framework/blob/alpha/src/lib/journey/stages/username-password.svelte#L117.

It's getting a bit late, so I'll pull the code down locally tomorrow morning and complete my functional test. It would be nice for future testing if we can move the social configuration from Stoyan's tenant to the one dedicated to this project. That way we can ensure PRs are testable with the automatically deployed code, reducing our need to pull it down locally.

Also, @ryanbas21, it seems Chromatic was entirely removed for our CI. Chromatic doesn't seem to have anything from 16 days ago on.

daveadams56:
> I reviewed the code, and it looks great. Unfortunately, I can't test the full feature with our usual Vercel deployment, so I'm have to pull this down locally and test it. I did find a few visual issues captured here: https://share.getcloudapp.com/4gu4J7gx.
> 
> The Google icon is a bit out of proportion, slightly misaligned and quite close to the text. The horizontal rule isn't as subtle as it is in light mode, so that should be adjusted. It looks like you're missing some of the classes on the `<hr />` element, seen here: https://github.com/cerebrl/forgerock-web-login-framework/blob/alpha/src/lib/journey/stages/username-password.svelte#L117.
> 
> It's getting a bit late, so I'll pull the code down locally tomorrow morning and complete my functional test. It would be nice for future testing if we can move the social configuration from Stoyan's tenant to the one dedicated to this project. That way we can ensure PRs are testable with the automatically deployed code, reducing our need to pull it down locally.
> 
> Also, @ryanbas21, it seems Chromatic was entirely removed for our CI. Chromatic doesn't seem to have anything from 16 days ago on.

@cerebrl thanks for catching the visual bugs - I have pushed a fix for the two things you mentioned. With regards to the social providers on your tenant for testing, I created a ticket for this ([SDKS-2249](https://bugster.forgerock.org/jira/browse/SDKS-2249)) so that we can prioritise the development of the social functionality before the holidays. Depending on how much other rework I need to do perhaps I can pull this one in today / tomorrow. Otherwise it will be something to do in the very near future, I agree. For your local testing, I will send on slack the values I used in layout.ts to point to Stoyan's tenant. Thanks!

tests/widget/modal/widget-modal.login-social-redirect.test.js
ryanbas21, line 9: when we pass over code like this, i think its a good idea to update the api to the newer style.

this should probably be `page.getByRole`

ryanbas21 (changes requested):

tests/widget/modal/widget-modal.login-social-redirect.test.js
daveadams56: Good point. It's weird, I thought I fixed references to this in the previous PR. Anyway I'll fix it now. Thanks.

cerebrl (approved):
There are some small improvements that I'd like to suggest, but I think I'll add them to one of the follow-up stories in our backlog. I think this is good for our first implementation.

ryanbas21 (approved):

PR #54, opened 2022-12-14
build(tsup and packaging for npm): this changes the build system to t…

ryanbas21:
…sup for packaging the widget

package the widget with tsup. its a builder that wraps esbuild and npm-dts. it should be able to generate the d.ts files however right now they generate as empty. it does build both cjs and esm builds. we keep the package.json file as is in the package/ folder.


.gitignore
```diff
@@ -1,4 +1,5 @@
 .DS_Store
+/dist
```
daveadams56, line 2: I'm curious - what within this project generates a `/dist` directory? The default build goes to `/package` doesn't it?

daveadams56 (approved):
This looks good to me. I reviewed the changes, pulled down the code and ran the new commands and all appears well. Just the one minor question but certainly nothing blocking. Approved.

.gitignore
ryanbas21: this was the default for tsup so i added it when i was  messing around.

package.json
```diff
@@ -15,13 +15,13 @@
     "dev": "npm run build:widget && npm run dev:widget | npm run dev:app",
     "dev:app": "vite dev",
     "dev:style": "postcss --watch src/lib/widget/main.css --output package/widget.css",
-    "dev:widget": "rollup -cw",
+    "dev:widget": "tsup --watch",
     "build": "npm run build:widget && npm run build:app",
     "build:app": "vite build",
     "build:storybook": "build-storybook",
     "build:widget": "rollup -c",
```
cerebrl, line 22: This was the intended `build` command to build the widget, rather than `package`. I feel like keeping this syntax helps keep the overall project stay consistent. Is there a strong reason why you want to shift to the `package` command?

package/package.json
```diff
@@ -3,8 +3,20 @@
...
+    },
+    "./modal": {
+      "import": "./modal.js",
+      "require": "./modal.cjs"
+    }
   },
-  "dependencies": {}
+  "dependencies": {
+    "@forgerock/javascript-sdk": "^3.4.0",
+    "esbuild": "^0.16.6",
```
cerebrl, line 17: Does this need to be listed in this `/package/package.json`?

tests/widget/modal/widget-modal.configurable-journey.test.js
```diff
@@ -23,10 +23,12 @@ test('modal widget and testing changing journeys', async ({ page }) => {
 
   // Change to go back to Sign In
   const backToSignInBtn = page.getByRole('link', { name: 'Sign in here!' });
+  await backToSignInBtn.waitFor({ state: 'visible' });
```
cerebrl, line 26: I'm really torn on this. In researching this, I believe any element we get and then apply some sort of action to it, like this:

```
const backToSignInBtn = page.getByRole('link', { name: 'Sign in here!' });
await backToSignInBtn.click(); <-- this should internally have a `waitFor` built into it
```

... has a `waitFor` built into it, so adding a `waitFor` duplicates this and is a bit misleading. I get why we want to add it, but let's keep iterating on this and make sure we really understand what we are adding here. I just say this in order to keep our tests from getting bloated without any real value being delivered to offset the cost.

If stuff is timing out because of the server, we _can_ add a custom timeout to the element query itself.

tests/widget/modal/widget-modal.configurable-journey.test.js
```diff
@@ -23,10 +23,12 @@ test('modal widget and testing changing journeys', async ({ page }) => {
...
   await backToSignInBtn.click();
 
   // Click "forgot password"
   const forgotPasswordBtn = page.getByRole('button', { name: 'Forgot password?' });
+  await forgotPasswordBtn.waitFor({ state: 'visible' });
```
cerebrl, line 31: Same with this.

package.json
```diff
@@ -31,12 +31,14 @@
   },
   "dependencies": {
     "@forgerock/javascript-sdk": "^3.4.0",
+    "esbuild": "^0.16.6",
```
cerebrl, line 34: I think this should be moved to a dev dependency, yeah?

tsup.config.js
```diff
@@ -0,0 +1,22 @@
...
+    entry: ['./src/widget/inline.svelte', './src/widget/modal.svelte'],
+    watch: false,
+    declaration: true,
+    minify: false,
+    plugins: [
+      postcssPlugin({
+        plugins: [autoprefixer],
+      }),
+    ],
+    sourcemap: false,
```
cerebrl, line 19: If we have `sourcemap` as false, are we preventing our ability to debug this code in the browser? I'd really like to preserve our ability to step through code in a meaningful way.

cerebrl (changes requested):
Can we get an example output of this PR for the package? Just a screenshot of the files and top-level dirs in the `package/` directory?

package.json
ryanbas21: I switched it so we had testing/ development against the build system that was being used to package. I can switch back if its preferred otherwise. wanted testing and stuff to go off the "package" version

package/package.json
ryanbas21: hmm, probaly not good catch.

tsup.config.js
```diff
@@ -0,0 +1,22 @@
+import autoprefixer from 'autoprefixer';
+import { defineConfig } from 'tsup';
+import postcssPlugin from '@chialab/esbuild-plugin-postcss';
+
+export default defineConfig(() => {
+  return {
+    clean: false,
+    dts: true,
+    format: ['esm', 'cjs'],
+    entry: ['./src/widget/inline.svelte', './src/widget/modal.svelte'],
```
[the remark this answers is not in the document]
ryanbas21: No, it doesn't bundle 3rd party dependencies from what i can tell. It does create two seperate bundles.

tsup.config.js
ryanbas21: I just havent had time to do any testing around sourcemaps in this version of hte story, so i left it as false for now.

package.json
ryanbas21: > Can we get an example output of this PR for the package? Just a screenshot of the files and top-level dirs in the `package/` directory?

Screen Shot 2022-12-15 at 1 12 19 PM

PR #55, opened 2023-01-03
feat(root): update-package-publishing

ryanbas21:
jira: https://bugster.forgerock.org/jira/browse/SDKS-2082

deliver a esm and cjs style bundle through rollup for package delivery via npm.
bundle types through svelte-kit & a type bundler.


rollup.config.js
```diff
@@ -1,104 +1,131 @@
...
+    },
 
-export default {
-  input: ['src/widget/modal.svelte', 'src/widget/inline.svelte'],
-  onwarn: (warning) => {
-    // TODO: Revisit once the JS SDK is updated to pure ES Modules
-    if (warning.code === 'THIS_IS_UNDEFINED') {
-      return;
-    }
+    emitCss: false,
```
cerebrl, line 32: May I ask why is this `false`?

rollup.types.config.js
```diff
@@ -0,0 +1,39 @@
+import dts from 'rollup-plugin-dts';
+import styles from 'rollup-plugin-styles';
```
cerebrl, line 2: I'm a bit confused by this config. I would think it's just about generating types, but then there's this `rollup-plugin-styles` module in it as well. Could you elaborate on this?

svelte.config.js
```diff
@@ -11,6 +11,10 @@ const config = {
   kit: {
     adapter: process.env.PREVIEW ? node() : auto(),
   },
+  package: {
+    dir: 'svelte-package',
+    extensions: ['.ts', '.svelte'],
```
cerebrl, line 16: Is this `.ts` item still needed here? I think this was one of our failed attempts at a fix.

tsconfig.json
```diff
@@ -2,20 +2,7 @@
...
-      "$lib": ["src/lib"],
-      "$lib/*": ["src/lib/*"],
-      "$locales": ["src/locales"],
-      "$locales/*": ["src/locales/*"],
-      "$package": ["package"],
-      "$package/*": ["package/*"],
-      "$widget": ["src/widget"],
-      "$widget/*": ["src/widget/*"]
-    },
+    "paths": {},
```
cerebrl, line 5: Bye 😢

cerebrl (commented):
This is looking good. Just left a few questions.

Dockerfile
```diff
@@ -12,7 +12,7 @@ COPY bashrc /root/.bashrc
 WORKDIR /usr/src/app
 
 COPY package.json package-lock.json* ./
-RUN npm install
+RUN npm install --legacy-peer-deps
```
ryanbas21, line 15: the svelte ecosystem doesnt seem very caught up, so this was the only solution i could find to make the preview work

rollup.config.js
cerebrl: Was this property supposed to be added to the `rollup.types.config.js`?

rollup.types.config.js
```diff
@@ -0,0 +1,36 @@
+import dts from 'rollup-plugin-dts';
+
+export default [
+  {
+    input: 'svelte-package/lib/widget/inline.svelte.d.ts',
+    external: ['./main.css'],
```
cerebrl, line 6: Is this so that Rollup doesn't try to find or include the types for the CSS file?

src/lib/constants.ts
```diff
@@ -18,7 +18,7 @@ declare global {
 
 export const AM_COOKIE_NAME = import.meta.env.VITE_FR_AM_COOKIE_NAME;
 export const AM_DOMAIN_PATH = import.meta.env.VITE_FR_AM_URL;
-export const AM_DOMAIN = extractDomainFromUrl(import.meta.env.VITE_FR_AM_URL);
+export const AM_DOMAIN = extractDomainFromUrl(import.meta.env.VITE_FR_AM_URL || "loginwidget.example.com");
```
cerebrl, line 21: Can we add a TODO comment here to re-evaluate this as it feels a bit like a hack without understanding the underlying issue here?

src/lib/journey/callbacks/confirmation/confirmation.story.svelte
```diff
@@ -3,7 +3,7 @@
   import Centered from '$components/primitives/box/centered.svelte';
 
   import Confirmation from './confirmation.svelte';
-  import type { CallbackMetadata, StepMetadata } from '$journey/journey.interfaces';
+  import type { CallbackMetadata, StepMetadata } from '../..//journey.interfaces';
```
cerebrl, line 6: Not sure why the build didn't catch this, but it looks like you have a double slash here.

src/lib/journey/callbacks/confirmation/confirmation.svelte
```diff
@@ -1,18 +1,18 @@
...
+  import { interpolate, textToKey } from '../../../_utilities/i18n.utilities';
 
   import type {
     CallbackMetadata,
     SelfSubmitFunction,
     StepMetadata,
-  } from '$journey/journey.interfaces';
-  import type { Style } from '$lib/style.store';
-  import type { Maybe } from '$lib/interfaces';
+  } from '../..//journey.interfaces';
```
cerebrl, line 13: Double slash here too.

src/lib/journey/callbacks/hidden-value/hidden-value.svelte
```diff
@@ -5,9 +5,9 @@
     CallbackMetadata,
     SelfSubmitFunction,
     StepMetadata,
-  } from '$journey/journey.interfaces';
-  import type { Style } from '$lib/style.store';
-  import type { Maybe } from '$lib/interfaces';
+  } from '../..//journey.interfaces';
```
cerebrl, line 8: Double slash.

src/lib/journey/callbacks/kba/kba-create.svelte
```diff
@@ -2,20 +2,20 @@
...
+  import LockIcon from '../../../components/icons/lock-icon.svelte';
 
   import type {
     CallbackMetadata,
     SelfSubmitFunction,
     StepMetadata,
-  } from '$journey/journey.interfaces';
-  import type { Style } from '$lib/style.store';
-  import type { Maybe } from '$lib/interfaces';
+  } from '../..//journey.interfaces';
```
cerebrl, line 16: Double

src/lib/journey/callbacks/password/password.svelte
```diff
@@ -13,9 +13,9 @@
     CallbackMetadata,
     SelfSubmitFunction,
     StepMetadata,
-  } from '$journey/journey.interfaces';
-  import type { Style } from '$lib/style.store';
-  import type { Maybe } from '$lib/interfaces';
+  } from '../..//journey.interfaces';
```
cerebrl, line 16: //

src/routes/(app)/+page.server.ts
```diff
@@ -1,9 +1,9 @@
 import type { RequestEvent } from '@sveltejs/kit';
 import type { PageServerLoad } from './$types';
-import type { stringsSchema } from '$lib/locale.store';
+import type { stringsSchema } from '../../lib/locale.store';
```
cerebrl, line 3: Can we move this internal module to the group below? I try to keep the external modules separate from the internal modules.

src/routes/(app)/+page.svelte
```diff
@@ -2,15 +2,15 @@
...
+  import Journey from '../../lib/journey/journey.svelte';
+  import { initialize as initializeJourney } from '../../lib/journey/journey.store';
+  import { initialize as initializeContent } from '../../lib/locale.store';
+  import { initialize as initializeOAuth, type OAuthStore } from '../../lib/oauth/oauth.store';
+  import { initialize as initializeUser, type UserStore } from '../../lib/user/user.store';
   import { Config, FRUser, SessionManager } from '@forgerock/javascript-sdk';
 
-  import type { JourneyStore } from '$journey/journey.interfaces';
+  import type { JourneyStore } from '../../lib/journey/journey.interfaces';
   import { goto } from '$app/navigation';
```
cerebrl, line 14: This too

src/routes/(app)/register/+page.server.ts
```diff
@@ -1,9 +1,9 @@
 import type { RequestEvent } from '@sveltejs/kit';
 import type { PageServerLoad } from './$types';
-import type { stringsSchema } from '$lib/locale.store';
+import type { stringsSchema } from '../../../lib/locale.store';
```
cerebrl, line 3: This too.

src/routes/api/locale/+server.ts
```diff
@@ -1,9 +1,9 @@
 import type { RequestEvent } from '@sveltejs/kit';
 import type { RequestHandler } from './$types';
-import type { stringsSchema } from '$lib/locale.store';
+import type { stringsSchema } from '../../../lib/locale.store';
```
cerebrl, line 3: ^^

tools/processTypes.mjs
```diff
@@ -0,0 +1,32 @@
+import { readFileSync, writeFileSync } from 'fs';
+import { dirname, join } from 'path';
+import { fileURLToPath } from 'url';
+
```
cerebrl, line 4: Can we add a comment explaining what this does and why? I think I get it, but we may forget or a new person might not have a clue.

rollup.config.js
ryanbas21: was mostl yme trying different things. It probably should be `true` since we use the postcss plugin.

rollup.types.config.js
ryanbas21: This was because the rollup was erroring on css files, however this is probabaly due to the fact that it was mapping to the actual js files. i can try to remove this now that we solved that problem.

cerebrl (approved):

PR #61, opened 2023-01-13
refactor(journey): moved step metadata related work into the journey store

cerebrl:
- Moved metadata function calls from the stage components into the journey.store file
- Improved types


src/lib/journey/journey.store.ts
```diff
@@ -132,9 +137,21 @@ export function initialize(initOptions?: StepOptions): JourneyStore {
...
+      if (stageAttribute && stageAttribute.includes('{')) {
+        try {
+          stageJson = JSON.parse(stageAttribute);
+        } catch (err) {
+          console.warn('Stage attribute value was not parsable');
+        }
+      }
+
+      const callbackMetadata = buildCallbackMetadata(nextStep, initCheckValidation(), stageJson);
+      const stepMetadata = buildStepMetadata(callbackMetadata, stageJson);
```
cerebrl, line 154: This is the logic for extracting the data out of a stage attribute and passing it to the metadata functions.

src/lib/journey/journey.store.ts
```diff
@@ -7,8 +7,11 @@ import type { JourneyStore, JourneyStoreValue, StackStore, StepTypes } from './j
 import { interpolate } from '$lib/_utilities/i18n.utilities';
 import {
   authIdTimeoutErrorCode,
+  initCheckValidation,
   shouldPopulateWithPreviousCallbacks,
-} from './_utilities/step.utilities';
+} from './stages/_utilities/step.utilities';
+import { buildCallbackMetadata, buildStepMetadata } from '$journey/_utilities/metadata.utilities';
```
cerebrl, line 13: It's worth noting that I pulled the responsibility of this metadata orchestration out of the stage component and into the store. I felt that keeping this responsibility in the stage component confused the purpose of a Svelte component as compared to a Svelte store.

Stores are where network calls are made and data is constructed. Svelte components should be focused on how that data gets rendered to the screen. Any state that they do control is just local component state. 

Moving all of this into to the store cleans up the stage components and focuses them more on rendering.

src/lib/journey/_utilities/callback-mapper.svelte
```diff
@@ -0,0 +1,210 @@
+<script lang="ts">
```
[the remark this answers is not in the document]
ryanbas21: I spent some time trying to actually update some of the SDK types for this exact issue with `getType`. I got pretty close but not where i wanted it to be. I think we could make a very quick fix to have `getType` return a `CallbackType` instead quite easily.

.svelte-kit/tsconfig.json
```diff
@@ -1,76 +1,42 @@
...
-		"../tests/**/*.ts",
-		"../tests/**/*.svelte"
-	],
-	"exclude": [
-		"../node_modules/**",
-		"./[!ambient.d.ts]**"
-	]
-}
\ No newline at end of file
+  "compilerOptions": {
```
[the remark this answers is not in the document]
ryanbas21: i've noticed this as well, i can take a look next sprint so we don't get this every single time we commit.

ryanbas21 (commented):
just a few notes while ive been reading through

src/lib/journey/_utilities/metadata.utilities.ts
```diff
@@ -1,67 +1,60 @@
...
 export function buildCallbackMetadata(
-  step: WidgetStep,
+  step: FRStep,
   checkValidation: (callback: FRCallback) => boolean,
+  stageJson?: Record<string, unknown> | null,
 ) {
+  const callbackCount: Record<string, number> = {};
+
   return step?.callbacks.map((callback, idx) => {
+    const cb = callback as FRCallback;
```
ryanbas21, line 27: I don't think the casting here is needed. the callbacks array is already of `FRCallback` so `cb` should have that type. Right?

ryanbas21 (commented):
I'm not 100% sure how i feel about the idea of specifically separating "derived" data and data thats retrieved from the server. On the surface, I understand the distinction, but i'm not understanding the value it provides.

Perhaps my mind is cloudy from years of redux, but why do we need to include the distinction in the metadata itself? Maybe i need more time to digest it, but i figured i'd just lay out some feelings i have through a few readings.

src/lib/journey/_utilities/metadata.utilities.ts
```diff
@@ -71,113 +64,41 @@ export function buildCallbackMetadata(
  * @param {array} callbackMetadataArray - The array returned from buildCallbackMetadata
  * @returns {object}
  */
-export function buildStepMetadata(callbackMetadataArray: CallbackMetadata[]) {
-  const numOfUserInputCbs = callbackMetadataArray.filter((cb) => !!cb.isUserInputRequired).length;
+export function buildStepMetadata(
+  callbackMetadataArray: CallbackMetadata[],
+  stageJson?: Record<string, unknown> | null,
```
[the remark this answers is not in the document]
cerebrl: If I can just swap out the generic for what I have without worry about it's affects downstream, then I can do it. If it requires additional code changes, I'll probably ask for us to write up a story for type improvements specifically.

src/lib/journey/_utilities/metadata.utilities.ts
```diff
@@ -71,113 +64,41 @@ export function buildCallbackMetadata(
...
-  if (userInputOptional) {
-    return true;
+  if (stageJson) {
+    stageMetadata = Object.keys(stageJson).reduce((prev, curr) => {
+      // Filter out objects or arrays as those are for the callbacks
+      if (typeof stageJson[curr] !== 'object') {
+        prev[curr] = stageJson[curr];
+      }
+      return prev;
+    }, {} as Record<string, unknown>);
```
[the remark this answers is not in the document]
cerebrl: Okay, I can try this, but I'm not sure what we gain from it. Can you explain?

src/lib/journey/_utilities/metadata.utilities.ts
```diff
@@ -71,113 +64,41 @@ export function buildCallbackMetadata(
  * @param {array} callbackMetadataArray - The array returned from buildCallbackMetadata
  * @returns {object}
  */
-export function buildStepMetadata(callbackMetadataArray: CallbackMetadata[]) {
-  const numOfUserInputCbs = callbackMetadataArray.filter((cb) => !!cb.isUserInputRequired).length;
+export function buildStepMetadata(
```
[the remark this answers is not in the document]
cerebrl: Will take a look

src/lib/journey/_utilities/metadata.utilities.ts
cerebrl: I can check.

src/lib/journey/callbacks/confirmation/confirmation.mock.ts
```diff
@@ -66,6 +66,37 @@ export default {
...
+          name: 'messageType',
+          value: 0,
+        },
+        {
+          name: 'options',
+          value: ['I confirm', 'No, I do not confirm'],
+        },
+        {
+          name: 'optionType',
+          value: -1,
```
daveadams56, line 86: It doesn't seem to affect what we're doing here, but out of interest do we know what this means?

src/lib/journey/callbacks/confirmation/confirmation.svelte
```diff
@@ -53,19 +50,24 @@
...
-    inputName = typedCallback?.payload?.input?.[0].name || `confirmation-${callbackMetadata.idx}`;
-    options = typedCallback.getOptions().map((option, index) => ({ value: `${index}`, text: option }));
-    defaultChoice = typedCallback.getDefaultOption();
+    inputName = callback?.payload?.input?.[0].name || `confirmation-${callbackMetadata?.idx}`;
+    options = callback.getOptions().map((option, index) => ({ value: `${index}`, text: option }));
+
+    if (callbackMetadata?.platform?.showOnlyPositiveAnswer) {
+      // The positive option is always first in the options array
+      options = options.slice(0, 1);
+    }
```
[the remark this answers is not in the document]
daveadams56: I agree there is no way to know for certain - I did a bit of digging on this and couldn't find anything.

src/lib/journey/callbacks/password/validated-create-password.stories.js
```diff
@@ -61,3 +70,30 @@ Interaction.play = async ({ canvasElement }) => {
   expect(text.type).toEqual('text');
   expect(cb.payload.input[0].value).toBe('myPassword');
 };
+
+export const ConfirmInteraction = Template.bind({});
+
+ConfirmInteraction.args = { ...ConfirmPassword.args };
+
+ConfirmInteraction.play = async ({ canvasElement }) => {
```
daveadams56, line 78: I had a test failure in this test when running `npm run test:storybook`:

```
Summary of all failing tests
 FAIL  src/lib/journey/callbacks/password/validated-create-password.stories.js (25.121 s)
  ● Callbacks/ValidatedCreatePassword › ConfirmInteraction › play-test

    page.evaluate: StorybookTestRunnerError: 
    An error occurred in the following story. Access the link for full output:
    http://localhost:6006/?path=/story/callbacks-validatedcreatepassword--confirm-interaction&addonPanel=storybook/interactions/panel

    Message:
     expect(received).toBe(expected) // Object.is equality

    Expected: "false"
    Received: "true"


    --------------------------------------------------

    Browser logs:

    warn: <Validated_create_password> was created with unknown prop 'stepMetadata'

    error: {"matcherResult":{"actual":"true","expected":"false","message":"expect(received).toBe(expected) // Object.is equality\n\nExpected: \"false\"\nReceived: \"true\"","name":"toBe","pass":false}}
```

tests/widget/inline/widget-inline.register.test.js
```diff
@@ -6,7 +6,7 @@ import { asyncEvents, verifyUserInfo } from '../../utilities/async-events.js';
 test('Inline widget with user registration', async ({ page }) => {
   const { clickButton, navigate } = asyncEvents(page);
 
-  await navigate('widget/inline?journey=Registration');
+  await navigate('widget/inline?journey=TEST_Registration');
```
daveadams56, line 9: I take it that TEST_ is the prefix for all of the journeys we are reserving for automated tests from now on?

daveadams56 (approved):
This looks great! I particularly like the typing of callbacks now throughout the widget, this makes things easier to read and work with. I focused a good bit of my effort on testing against the server and found that what you've implemented copes with the different combinations nicely. I also checked out and ran tests - I just had one failure which I've mentioned in a comment. Other than that, my comments are non-blocking / just questions or praising the great work here! I'm happy to approve this, assuming you can look into / are happy with the storybook test failure.

src/lib/journey/callbacks/password/validated-create-password.stories.js
cerebrl: Ah, that was an interesting bug. It only failed when the command is run, but not when I test is run manually in the UI. Thanks for catching this.

tests/widget/inline/widget-inline.register.test.js
cerebrl: That is correct. I accidentally broke some tests because I was altering a journey that Ryan was using for e2e, so I figured it would be better to have dedicated journeys for testing.

cerebrl:
All comments should be addressed and responded to.

PR #62, opened 2023-01-17
SDKS 2268/textoutput button render

ryanbas21:
jira: https://bugster.forgerock.org/jira/browse/SDKS-2268

add button when a textoutput node has no button to continue the journey

### DIFF
--- a/src/lib/journey/callbacks/text-output/text-output.svelte
+++ b/src/lib/journey/callbacks/text-output/text-output.svelte
@@ -3,8 +3,6 @@
   import sanitize from 'xss';
 
   import Text from '$components/primitives/text/text.svelte';
-  import Button  from '$components/primitives/button/button.svelte';
-  import T from '$components/_utilities/locale-strings.svelte';
 
   import type {
     CallbackMetadata,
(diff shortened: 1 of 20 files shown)


src/lib/journey/callbacks/text-output/text-output.svelte
```diff
@@ -15,23 +17,31 @@
...
     cleanMessage = sanitize(dirtyMessage);
+    shouldRenderTextOutputNext = stepMetadata !== null && (stepMetadata as StepMetadata).isStepSelfSubmittable && !(stepMetadata as StepMetadata).isUserInputOptional
   }
 </script>
 
 <Text classes="tw_font-bold tw_mt-6">
   {@html cleanMessage}
 </Text>
+
+{#if shouldRenderTextOutputNext}
```
cerebrl, line 42: I'm really hesitant to have a callback component render the "Next" button. I say that because the complex compositional nature of callbacks. One may use the `TextOutputCallback` along with other callbacks. This may result in two, form-submitting buttons being rendered, like this paired with a `ConfirmationCallback`. 

What's more, I wouldn't want this callback to bear the responsibility of it needing awareness of other callbacks, so I think we should have only one place where the "Next" button is rendered and that's at the stage level. 

When I suggested that the stage component shouldn't have to explicitly need to know about callbacks, I didn't necessarily mean that the responsibility should be delegated to the callback component. My suggestion is we might need to improve the generation of the metadata itself to include this use case as a state with which the "Next" button is conditionally rendered in the stage component.

If this doesn't make sense. We can chat about it on Slack or video call. Let me know.

cerebrl (changes requested):

daveadams56 (commented):
I wanted to post a partial review covering my thoughts on how to simplify the change and keep the metadata clean - however there is a problem that we definitely need to solve here which is how to make this work for TextOutput (with next button), and SuspendedTextOutput (without next button). I'm still thinking on this one.

PR #64, opened 2023-01-24
SDKS-2337/link-bug-requirement

ryanbas21:
Reported bug: https://bugster.forgerock.org/jira/browse/SDKS-2337

The link was required which breaks the modal via zod error.

### DIFF
--- a/src/lib/widget/inline.svelte
+++ b/src/lib/widget/inline.svelte
@@ -183,7 +183,7 @@
   export let config: z.infer<typeof partialConfigSchema>;
   export let content: z.infer<typeof partialStringsSchema>;
   export let journeys: z.infer<typeof journeyConfigSchema> | undefined = undefined;
-  export let links: z.infer<typeof partialLinksSchema>;
+  export let links: z.infer<typeof partialLinksSchema> = {};
   export let style: Style | undefined = undefined;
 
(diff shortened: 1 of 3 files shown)


src/lib/widget/inline.svelte
```diff
@@ -183,7 +183,7 @@
   export let config: z.infer<typeof partialConfigSchema>;
   export let content: z.infer<typeof partialStringsSchema>;
   export let journeys: z.infer<typeof journeyConfigSchema> | undefined = undefined;
-  export let links: z.infer<typeof partialLinksSchema>;
+  export let links: z.infer<typeof partialLinksSchema> = {};
```
cerebrl, line 186: Could you do me a favor and, rather than added an empty object, could you add `.optional()` to the Zod type here (reflective of the fact that it really is optional): https://github.com/cerebrl/forgerock-web-login-framework/blob/main/src/lib/links.store.ts#L6? I think this would be a bit cleaner and keep the fix right at the source.

Example:

```
// links.store.ts
export const linksSchema = z.object({
  termsAndConditions: z.string(),
}).optional();
```

Zod documentation: https://zod.dev/?id=optional

cerebrl (changes requested):

cerebrl (approved):

PR #66, opened 2023-01-30
sdks 2057/checkmarx compiled code

ryanbas21:
https://bugster.forgerock.org/jira/browse/SDKS-2057

### DIFF
--- a/tools/modify-gitignore.sh
+++ b/tools/modify-gitignore.sh
@@ -1,0 +1,21 @@
+# shell script to modify the gitignore file during ci
+
+# We want to commit back the build package after a release
+
+# so we can run checkmarx which is behind a vpn and doesn't
+
+# support svelte out of the box. This file is just
+
+# to add to the bottom of the gitignore during our release
+
+# a !package/\* so that we unignore packages but we don't commit
+
+# this change back. So locally we as developers are able to
+
+# still ignore the package, and are not responsible for cluttering
+
+# our pull requests with build rollup files.
+
+# see the releaserc file and exec command for where this file is run.
+
+echo '!package/\*' >> .gitignore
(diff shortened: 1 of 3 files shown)


tools/modify-gitignore.sh
```diff
@@ -0,0 +1,21 @@
...
+
+# this change back. So locally we as developers are able to
+
+# still ignore the package, and are not responsible for cluttering
+
+# our pull requests with build rollup files.
+
+# see the releaserc file and exec command for where this file is run.
+
+echo '!package/\*' >> .gitignore
```
cerebrl, line 21: This make sense. I'm assuming this PR only contains this one change to the CI, but does not contain the other changes regarding how the files are built and where they are persisted, yeah?

tools/modify-gitignore.sh
```diff
@@ -0,0 +1,21 @@
+# shell script to modify the gitignore file during ci
+
+# We want to commit back the build package after a release
```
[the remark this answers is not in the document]
ryanbas21: It's not included in the `assets` array of the semantic-release/git` plugin, which is committing the assets back.

tools/modify-gitignore.sh
cerebrl: The important thing for me is that we don't bundle the JS that's scanned. We just compile the `.svelte` files to JS, and leave them entirely modular and preserve the directory structure.

The reason for this is so when we scan the files and it finds something, it reports that issue from the exact file and directory structure. That way we can immediately investigate the issue. 

If we bundle all the modules, then it could report an issue on line 2198 on `widget.js` (or worse, on line 1 because it's also minified). Trying to reverse engineer where that issue is in the source code might be a huge timesuck.

cerebrl (approved):
Okay, let's try it and see how the scan reports the issues.

PR #69, opened 2023-01-30
fix(update-rollup): es only locally, cjs in CI for release

ryanbas21:
don't build cjs locally unless a CI=true is set which is always set in CI!

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -88,6 +88,7 @@ function warningHandler(warning) {
   // console.warn everything else
   console.warn(warning.message);
 }
+const { CI = false } = process.env;
 
 export default [
   /** ****************************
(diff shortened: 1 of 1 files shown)


rollup.config.js
```diff
@@ -119,26 +120,30 @@ export default [
...
+        output: {
+          file: path.join(packageFolder, 'modal.cjs'),
+          format: 'cjs',
+          sourcemap: true,
+          inlineDynamicImports: true,
+        },
+        plugins,
+      }
+    : undefined,
+].filter(Boolean);
```
cerebrl, line 149: This syntax always trips me up when I see it. I'm not a huge fan of this pattern, but I think it's fine for this context.

cerebrl (approved):

